PR #182, opened 2014-12-24
Added config options for disabling weak rituals

LucidTheStick:


bookerthegeek:
Any idea on this request?  Is it good to go or does it need something more?

PR #210, opened 2015-01-13
Well of suffering ritual with multiple target altars behavior

jco2641:
Well of suffering ritual should deliver blood to the first altar it finds within its range, rather than the last.


nabijaczleweli:
The `&& !testFlag` is after the `if(...)` brackets, isn't it?

Issue #12, opened 2014-04-04
[Enhancement] Localize chat messages


Vexatos:
Currently, only block/item names as well as the Blood Altar NEI entry are localized. There are still many strings not localized, especially the chat messages in this mod.

Adaptivity:
I'm done it! :smile:

PR #220, opened 2015-01-22
Long-awaited feature(Localized messages and more)

Adaptivity:
Finally! Months of waiting. And here it is: localized everything. Should finally close: #12
Don't merge yet. Server crashing at startup. Solution: adding @SideOnly(Side.CLIENT) doesn't help.


Adaptivity:
https://github.com/Adaptivity/BloodMagic/pull/1 this can add localization support for rituals. Waiting for your opinion @WayofTime

PR #234, opened 2015-01-30
Add Teleposer blacklist

TehNut:
Allows blacklisting by using the `modid:blockname:meta` format.

Example:

`minecraft:stone:0` will stop Vanilla stone from being teleposed.


TehNut:
Moved to #259

PR #237, opened 2015-01-31
Changes to Item Binding

hilburn:
Allows event to cancel the binding (rather than set key to "" which had the same effect but was messy)

Prevents binding by Fake Players


WayofTime:
The reason that I have the items act even if they were not bound is because I'd want them to harm the player otherwise. Will possibly merge later, after thinking on what'd I'd want to do.

hilburn:
Ah I see. Didn't it always bind the item though, so that situation never really occurred? That was my reading of it when I was looking at possible knock-on effects.

It would be relatively simple to split up the modified thing to allow different behaviour in the event of a failed binding, so lemme know

WayofTime:
The native FakePlayer handling will probably be enough for me to accept the changes anyway. Guess it makes more sense for the items to not work if they don't taste the first few drops of life.

PR #242, opened 2015-02-02
Account for Use Golems in the fake player check

Vazkii:


WolfAmaril:
Wouldn't use golems be mostly useless for most blood magic uses, as their speed of use would wind up with them dead long before anything useful was done?

Vazkii:
Might as well check for them anyway, won't hurt ._.

WolfAmaril:
Fair 'nuff.

shadowfacts:
You forgot the closing ]

Vazkii:
If the ] was there it wouldn't work.

WayofTime:
Thanks, Vazkii! Wish people would actually use FakePlayer, but alas~

Vazkii:
The problem with using the forge fake player is that it's an EntityPlayerMP, for all the problems that can cause.

WayofTime:
Boo. Oh well, at least this can be further expanded later. Thanks again.

PR #243, opened 2015-02-02
Added Achievements

Arcaratus:
Added the frame for achievements. You still need to add them (or have Madman do it) via ModAchievements and updating the en_US.lang


WayofTime:
Sure... I guess?

PR #260, opened 2015-02-20
Fix teleposerBlacklist shorthand support

TehNut:
I derped and forgot to support `modid:blockname` automatically using a meta of `0`.


Alex-hawks:
found a bug in that code. you want to do 
```Java
((block == event.initialBlock && ( meta == event.initialMetadata || meta == OreDictionary.WILDCARD_VALUE )) || (block == event.finalBlock && ( meta == event.finalMetadata || meta == OreDictionary.WILDCARD_VALUE))) 
```
at line 702

PR #272, opened 2015-03-07
Only reflect Projectiles that fly in the direction of the player

MaPePeR:
I tested this by setting `throwingEntity` to `null` and then firing Vanilla Arrows around. 

The change allowed me to shoot arrows then (except when aiming very low to the ground).

Sadly my Computer does not appear to have enough memory to handle 2 Clients and a Server, so i could not really test if it still deflects all the incoming projectiles. (It did work with skeletons, though)


Alex-hawks:
use a LAN Server

MaPePeR:
Not sure, if I can manage 2 clients, but I will try.

MaPePeR:
I tried it with "Open to LAN" now and it still worked as expected and reflected incoming projectiles.

WayofTime:
Sorry I'm late, lots o' stuff happening lately.

WayofTime:
Oops!~

PR #284, opened 2015-03-24
Should fix(? someone check) a minor bug with imperfect ritual stones

Arcaratus:
There was a bug involving Thaumcraft golems being able to activate imperfect ritual stones without usage of LP, this -->might fix it, don't know for sure, someone has to check and see.

--added a isFakePlayer check


WayofTime:
Already corrected this, but forgot to push the fix :P

PR #300, opened 2015-04-13
Implement reagent usage for well of suffering

Iamtk421:
Hi,
I'm not really sure of the process of submitting pull requests etc as this is the first time I've ever created one for any project., but would like to suggest this for discussion/consideration.

I was annoyed that reagents couldn't be used to increase the LP gain of the well of suffering, nor increase its range so implemented both using offensa and potentia, as well as lowered LP costs using sanctus.

### DIFF
--- a/src/main/java/WayofTime/alchemicalWizardry/common/rituals/RitualEffectWellOfSuffering.java
+++ b/src/main/java/WayofTime/alchemicalWizardry/common/rituals/RitualEffectWellOfSuffering.java
@@ -20,9 +20,9 @@ public class RitualEffectWellOfSuffering extends RitualEffect
 {
     public static final int timeDelay = 25;
     public static final int amount = 10;
-    public static final int offensaDrain =5;  LP amount per damage
-    public static final int potentiaDrain =10;  Increase Range
-    public static final int sanctusDrain = 5; Reduce LP use
+    public static final int offensaDrain =5; // LP amount per damage
+    public static final int potentiaDrain =10; // Increase Range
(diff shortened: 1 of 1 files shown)


WayofTime:
Had a similar thought, but haven't submitted it to git! Will look at this fully when I get home and decide which of them to use ;3

Iamtk421:
I'm glad you decided to put the increased LP return in. :-)
The well of suffering was my first foray into powering rituals with reagents, only to find after trying all of them that it didn't do anything.

I'm curious, if you don't mind, why did you decide to not implement increased range? It was one of the reasons I originally thought of using reagents hoping I'd be able to have a third spawning floor beneath my ritual and not have to do any silly mob transport. Of course I could have just spaced them better too.

It'd be good if the ritual stopped running, or at least stopped using reagent if the altar is full though.

TheMadmanofChaos:
The reason that "The ritual stops functioning when the altar is full" was not implemented was due to how most well of suffering mob farms run. If the ritual stops, the mobs build up, the entity count spirals out of control, the server starts to lag...

We did consider adding that effect as a separate reagent effect, however.

Iamtk421:
Yea, I kind of assumed that.

Now that WayofTime implemented reagent for it, perhaps it would be more appropriate that when full the ritual simply stops using reagent for increased LP.

WolfAmaril:
I'd say that instead of stopping reagent use at full, maybe stop it at 90%, or 95%. My altar is never really "Full" because I keep a blood orb in it at all time to keep constant power going to all my rituals, so I can see that leading to a lot of problems for the end users, if not for the server itself.

PR #319, opened 2015-05-06
Remove non-essential files from `../gradle/`

TehNut:
They got added in #306


TehNut:
Oh, I also upgradled to `2.3`

PR #348, opened 2015-06-10
Change ore block detection for Ritual of Magnetism

### LINKED ISSUE #347: Ritual of Magnetism not picking up Redstone Ore
I have made a contraption that uses Botania's Orechids, Autonomous Activators, Ritual of Magnetism and Ritual of the Crusher to generate Ores and harvest them automatically.
However some of the Redstone Ore is not being picked up by the Ritual of Magnetism. Any other ore that spawned previously to Redstone Ore at the same location was picked up just fine, Redstone Ores that spawn at other location are also picked up.

Forge: 1403
Blood Magic: 1.3.3-4

ljfa-ag:
Switched getOreID to getOreIDs to avoid problems with blocks that are oredicted more than once.
Also, added a shortcut for vanilla ores and subclasses.
That should fix #347.

### DIFF
--- a/src/main/java/WayofTime/alchemicalWizardry/common/rituals/RitualEffectMagnetic.java
+++ b/src/main/java/WayofTime/alchemicalWizardry/common/rituals/RitualEffectMagnetic.java
@@ -4,6 +4,8 @@ import java.util.ArrayList;
 import java.util.List;
 
 import net.minecraft.block.Block;
+import net.minecraft.block.BlockOre;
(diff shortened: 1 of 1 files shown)


ljfa-ag:
I just read this post by LexManos: https://github.com/MinecraftForge/MinecraftForge/pull/1926#issuecomment-110486973

The way ores are detected is pretty expensive indeed (and it wasn't that much better before the PR). I guess I'll make it so that the results of isBlockOre get cached.

PR #349, opened 2015-06-12
Small stuff

Arcaratus:
Changed some things


Arcaratus:
Realized what I was going to actually do...

PR #350, opened 2015-06-13
Added useful things (Guide-API woot!)

Arcaratus:
Added in an auto-downloader for the latest version of Guide-API for all those dopes like me who don't read the download section completly
Sigil of Magnetism now attracts XP orbs
Added the fillMax command which fills a player's SN to their current orb's max
Random things


TehNut:
Ew deploader

Arcaratus:
Neh

WayofTime:
I don't want people to be required to download Guide-API. It seems this PR makes it a dependency, and forces you to download it. I don't like it.

Arcaratus:
I'll make it a download chat option...

PR #352, opened 2015-06-13
Check if Tehnut's lying....

Arcaratus:
Did what he told me to do


TehNut:
i don't lie. i merely tell an incorrect truth

PR #354, opened 2015-06-14
Things

### LINKED ISSUE #342: Can't put lava sigil into filled socket
For some reason I can no longer put lava sigil's into filled sockets when making bound armour, it just uses the sigil and places lava down beside the socket rather than putting the sigil inside. I am using 1.3.3-2 so if this was fixed in the past two versions just let me know and I'll update

Arcaratus:
(I need to get better at naming these PR's)
-Added in the ritual dismantler
-Added the /soulnetwork <player> create command that creates a player's Soul Network
-Fixed #342 (Lava and Water sigils)


Arcaratus:
Recheck this PR please (able to merge)

PR #397, opened 2015-07-25
Ritual of Growth now supports base block crops

Timeslice42:
Modified the ritual of the green grove to support crops which need a special block under the soil to grow.

### DIFF
--- a/src/main/java/WayofTime/alchemicalWizardry/common/rituals/RitualEffectGrowth.java
+++ b/src/main/java/WayofTime/alchemicalWizardry/common/rituals/RitualEffectGrowth.java
@@ -60,6 +60,12 @@ public class RitualEffectGrowth extends RitualEffect
                             {
                                 this.canDrainReagent(ritualStone, ReagentRegistry.aquasalusReagent, aquasalusDrain, true);
                             }
+                            
+                            // Search one block higher for crops which need a specific block under the soil to grow
+                            if (SpellHelper.hydrateSoil(world, x + i, y + 2, z + j))
+                            {
+                                this.canDrainReagent(ritualStone, ReagentRegistry.aquasalusReagent, aquasalusDrain, true);
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


WayofTime:
Eh... most plants that require special blocks can use two blocks under it instead of just 1. Also, if you were to do this a for loop is a LOT more preferential than copy-paste. Also also, this effectively doubles its power...

Timeslice42:
Sentance 1) I'm not sure what you're referring to, but, for example, agricraft ore crops need the corresponding ore block underneath the soil block to grow at all. While you can certainly put a RotGG underneath the ore blocks, it doesn't reach the crops, hence the change.

Sentance 2) I was unsure exactly how the reagent costs were being calculated, and I didn't want it to ever be boosting crops for "free". Feedback on this is appreciated.

Sentance 3) The only case I can see this doubling this power is on crops that grow more than one block tall. Since this still wouldn't be for free I wasn't super-concerned about it. In the case of cactus and sugarcane, i'm pretty sure that only growth ticks on the top block can grow the plant anyways. Again, feedback on this is appreciated.

I'm happy to rewrite this based on your next comment.

Blue64:
I know you are correct on the Sugarcane, as I actually watched it grow for a while once (I was very bored and needed books more than ores lol), only the top block actually "grows", the one beneath it reverts to 0 when it realizes that the block above it is unable to be grown into.

Some plants require End Stone to be planted on (eg. Extra Utilities Ender Lilly), so this would be a useful patch (I hope). Would this have a compound effect with things such as Fertilized Dirt (i forgot which mod added it, but it basically speeds up the GrowthRate by simulating ```Config Value``` ticks for each tick the Dirt receives)?

Timeslice42:
This patch is actually for crops that need something *under* the block they are planted on.

Blue64:
Ah, I had misunderstood that, thanks for clearing that up, and for telling me the Ender Lillies are mostly hardcoded based upon their planted plot, I shall utilize that information well.

Timeslice42:
The math is here: http://ftbwiki.org/Ender-Lily_Seeds

Blue64:
[```"Ender-Lily Seeds can only be crafted in the Agrarian Skies Mod Pack"```](http://ftbwiki.org/Ender-Lily_Seeds#cite_ref-pack_1-0)

not anymore! ```*adds recipe to MineTweaker Scripts* :P``` that's _obviously_ the ```Ex Nihilo Grass Seeds``` lol I don't know which other mods even add ```Grass Seeds```, but I guess on F. T. B. servers, their options are limited. Thanks for the info @Timeslice42 (I need to upgrade from End Stone lol)

WayofTime:
Because 1.7.10 is no longer being worked on, and because 1.8.8 is nearing a beta release for the mod, I am closing this PR. If you wish to make a similar adjustment when we finally release the mod, feel free! :D

PR #416, opened 2015-08-02
Create a utility for registering renders for Blocks/Items

TehNut:
Almost direct copypasta from [my project](https://github.com/TehNut/ResourcefulCrops/blob/1.8/src/main/java/tehnut/resourceful/crops/util/helper/RenderHelper.java). Allows you to simply register models for Items/Blocks for given Metadata's.

Call these from your ClientProxy in postInit. A working example can be seen [here](https://github.com/TehNut/ResourcefulCrops/blob/1.8/src/main/java/tehnut/resourceful/crops/proxy/ClientProxy.java#L21-L25) [[Items](https://github.com/TehNut/ResourcefulCrops/blob/1.8/src/main/java/tehnut/resourceful/crops/registry/ItemRegistry.java#L42-L52)] [[Blocks](https://github.com/TehNut/ResourcefulCrops/blob/1.8/src/main/java/tehnut/resourceful/crops/registry/BlockRegistry.java#L30-L34)].

Review this a bit more closely. I did it from the web editor and it may not have worked 100% correctly.

### DIFF
--- a/src/main/java/WayofTime/alchemicalWizardry/client/InventoryRenderHelper.java
+++ b/src/main/java/WayofTime/alchemicalWizardry/client/InventoryRenderHelper.java
@@ -1,0 +1,79 @@
+package WayofTime.alchemicalWizardry.client;
+
+import net.minecraft.block.Block;
+import net.minecraft.client.Minecraft;
+import net.minecraft.client.renderer.ItemMeshDefinition;
+import net.minecraft.client.renderer.entity.RenderItem;
+import net.minecraft.client.resources.model.ModelBakery;
+import net.minecraft.client.resources.model.ModelResourceLocation;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemBlock;
+import net.minecraft.item.ItemStack;
+
+public class InventoryRenderHelper {
+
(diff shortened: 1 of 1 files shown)


Arcaratus:
I'm pretty sure @WayofTime is still working on something for it but he should take a look at that for references

TehNut:
This has damn near everything you need, tbh. There really isn't much else to build on it

WayofTime:
I think that this could work very well. Should have everything that we need, and for those blocks and items (mainly the items) that would require a special method for registering I can just make a simple loop for it. I'll keep it open for another day to discuss, but I think I'll accept this later.

TehNut:
@WayofTime Out of boredom, I began a fork over [here](https://github.com/TehNut/BloodMagic/tree/bloodmagic1.8) to play with 1.8 stuff. It is currently using this class [here](https://github.com/TehNut/BloodMagic/blob/bloodmagic1.8/src/main/java/WayofTime/alchemicalWizardry/ModItems.java#L310-L359).

WayofTime:
I'll accept it for now :3 It's too bad I'm visiting my sister and don't have access to my code, otherwise I'd be doing what you are doing right now, @TehNut.

PR #423, opened 2015-08-11
Vacation commit

### LINKED ISSUE #422: Ritual of Containment range upgrade reagent has no effect
As per the title. The forum post claims Aether extends the range, but this has no effect. According to the source file (https://github.com/WayofTime/BloodMagic/blob/master/src/main/java/WayofTime/alchemicalWizardry/common/rituals/RitualEffectContainment.java), the actual reagent to use is Terrae, but also according to that file, it still has no effect (flag hasTerrae is never used, and entity search range d0 is always 5 blocks).

Arcaratus:
Updated build.properties
Fixes #422 --check if you like the fix
Tweaked the Unbinding Recipe API just a bit
!!!!Check if I accidentally messed up anything with branches!!!


src/main/java/WayofTime/alchemicalWizardry/common/rituals/RitualEffectUnbinding.java
```diff
@@ -101,7 +101,20 @@ else if (UnbindingRegistry.isRequiredItemValid(itemStack))
                     item.setDead();
                     doLightning(world, x, y, z);
                     ItemStack[] inv = ((BoundArmour) itemStack.getItem()).getInternalInventory(itemStack);
-
+                    int bloodSockets = 0;
```
UndeadZeratul: Instead of chaining if/else-ifs, why not handle this via a switch/case?

If not that, what about storing the amount of sockets needed in a map keyed by the ModItems.BOUND_ITEM?

Basically, you'd then only have to do the following:

```
Map<Item, int> socketAmounts = buildSocketAmounts();

// ...

bloodSockets = socketAmounts.get(itemStack.getItem());
```

You'd basically just want to insert the number of sockets needed into the map, keyed by the item.  If there's no match, just set bloodSockets to 0.

Arcaratus: Since the Items are not final, switches would not work in this case.

UndeadZeratul: Ah, fair enough.

iambob314:
The fix for Terrae now extending Ritual of Containment (radius 5->10 blocks) looks great, thanks!

Issue #420, opened 2015-08-08
Bucket handler always returns iron bucket


DarkGuardsman:
https://github.com/WayofTime/BloodMagic/blob/master/src/main/java/WayofTime/alchemicalWizardry/common/LifeBucketHandler.java#L18

It seems the fill bucket event is not checking to ensure the bucket being passed into the event is an iron bucket. Instead it just returns a filled bucket causes the player to get a free iron bucket out of a non-iron bucket.

WayofTime:
Can you give an instance of this in-game?

Arcaratus:
Probably when using Ex Nihilo etc.

DarkGuardsman:
An example of this is a wooden bucket I coded up for a mod. In build 1 it would turned into an iron bucket of some of blood magic's fluid. Current build no longer has this issue as it seems most mods assume the ItemStack inputed into the FillBucketEvent is an iron bucket. So I have removed the forge event however it is still a bug that could effect other mods. 

Here is a video of what happens https://youtu.be/ilg5-Ln-l_0

Zsashas:
Also the Iguana Tweaks Clay Bucket may be affected.

PR #464, opened 2015-10-07
Various fixes and improvements (Details in description)

### LINKED ISSUE #421: Teleposer Blacklist
When i try to blacklist the Thaumcraft Auranodes from being teleposed (Thaumcraft:blockAiry:0) it works as expected but nitor (Thaumcraft:blockAiry:1) gets also blacklisted.
(when blacklisting nitor only nitor is blacklisted)
So i guess Metadata 0 is treated as every instance of the block, which should be changed, considering that leaving out the metadata is also blacklisting the whole block?

### LINKED ISSUE #441: Altar z-fights when it is empty
https://github.com/WayofTime/BloodMagic/blob/master/src/main/java/WayofTime/alchemicalWizardry/common/renderer/model/ModelBloodAltar.java#L53

Check that's greater than 0 before drawing anything...it's pointless to draw anything anyways.

### LINKED ISSUE #459: Blood altar doesn't work with chiseled beacons
I was trying to make a more aesthetically pleasing blood altar (using different beacons that give off different light, but apparently only the vanilla beacon works to make a tier 5 blood altar.

It'd be nice if chiseled beacons could work as well, since they're functionally equivalent and require the actual beacons to make anyway. :)

Edit: Also of note, probably caused by the same thing, are that chiseled glowstone and bloodstone also reduce the altar level, as they're not seen as valid blocks.

Crossposted at Chisel-2/Chisel-2#415 since i'm not sure which mod is at fault.

TehNut:
* Allow the altar to accept any instanceof BlockBeacon. Fixes #459 (Chisel-2/Chisel-2#415)
* Allow disabling of Weak Rituals. Resubmission of #182
* Fixes the Z-fighting issue when a Blood Altar was empty. It would still render blood at the same level as the bottom of the basin. Fixes #441
* Attempted fix at getting free Iron Buckets when non-Iron Buckets were used. Basically just disallows all non-Iron Buckets. Fixes (?) #420 (hehe)
* Drastically improve the Teleposer Blacklist feature. In this new implementation, wildcards are not accepted. I'll attempt to fix that soon. Fixes #421.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,6 +44,10 @@ repositories {
     	name 'TehNut Repo'
     	url 'http://tehnut.info/maven/'
     }
+    maven {
+        name 'tterrag Repo'
+        url 'http://maven.tterrag.com/'
+    }
 	ivy {
 		name "MineTweaker3"
         artifactPattern "http://minetweaker3.powerofbytes.com/download/[module]-[revision].[ext]"
@@ -57,6 +61,7 @@ dependencies {
 	compile name: 'MineTweaker3', version: config.minetweaker_version, ext: 'jar'
 	compile "codechicken:ForgeMultipart:1.7.10-1.1.0.314:dev"
 	compile "info.amerifrance.guideapi:Guide-API:" + config.mc_version + "-" + config.guideapi_version + ":deobf"
+    compile "team.chisel:Chisel:" + config.chisel_version + ":api"
 }
 
 minecraft {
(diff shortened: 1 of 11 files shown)


Drullkus:
*I'm the Shizzle artist and I approve of this.*

TehNut:
>**Shizzel**

TehNut:
Nope nope nope. don't merge this yet.

TehNut:
Safe again

TehNut:
Now 100% more chiselable

Drullkus:
> Now 100% more shizzelable

Yes, yes. Indeed

TheCricket:
10/10 Would Chisel again

PR #465, opened 2015-10-08
Fix Invisible Items

Necr0:
When the clients graphics are set to be "fast" the client renders items as 2d plains always facing the player and not as rotating 3d objects. So if the item rotates it is invisible half of the time because it has no backside. I don't know where else you used this code but please consider fixing it.
-Necr0


WayofTime:
Missing a }, dood.

WayofTime:
Oh, nvm, you used a method that didn't need them.

PR #491, opened 2015-11-03
Check what I should fix

Arcaratus:


TehNut:
Remove the changes to `AltarRecipeRegistry` and `NetworkHelper` as well as run it through the formatter, then I will merge.

PR #523, opened 2015-12-29
Fix #347 again...

### LINKED ISSUE #347: Ritual of Magnetism not picking up Redstone Ore

ljfa-ag:
When I made pull request #359 I inadvertently broke this method again for the case of lit redstone ore, since it has no item attached to it. This should work better now ^^

### DIFF
--- a/src/main/java/WayofTime/alchemicalWizardry/common/rituals/RitualEffectMagnetic.java
+++ b/src/main/java/WayofTime/alchemicalWizardry/common/rituals/RitualEffectMagnetic.java
@@ -33,11 +33,12 @@ public class RitualEffectMagnetic extends RitualEffect
     
     public static boolean isBlockOre(Block block, int meta)
     {
-        if (block == null || Item.getItemFromBlock(block) == null)
-            return false;
-        
+        //Special case for lit redstone ore
         if (block instanceof BlockOre || block instanceof BlockRedstoneOre)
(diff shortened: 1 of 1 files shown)


TehNut:
1.7.10 dev is complete. This isn't gamebreaking enough to warrant a new version.

Blue64:
Aw! bummer!

ljfa-ag:
Hm well...I guess it doesn't hurt to have this in anyway when you want to port it to 1.8, or when for some reason a new 1.7 version does get released.

WayofTime:
I'll merge this for now, even though I'm not sure we'll touch 1.7.10 again. Thanks!

Adaptivity:
If people actually want they can build the mod from the source. So, you made the right decision about merging this.

Blue64:
For those wondering how to do that, read the Forge GitHub ReadMe, it has instructions on that.

PR #527, opened 2015-12-31
Jumping Ritual

Arcaratus:
-Added in the semi-functional jumping ritual
-Some teleposer stuff


src/main/java/WayofTime/bloodmagic/item/ItemTelepositionFocus.java
```diff
@@ -67,22 +76,24 @@ public void addInformation(ItemStack stack, EntityPlayer player, List<String> to
...
 
     public World getWorld(ItemStack stack)
     {
-        stack = NBTHelper.checkNBT(stack);
         return DimensionManager.getWorld(stack.getTagCompound().getInteger(Constants.NBT.DIMENSION_ID));
     }
 
     public BlockPos getBlockPos(ItemStack stack)
     {
-        stack = NBTHelper.checkNBT(stack);
```
TehNut: why are you removing these checks...

PR #531, opened 2016-01-01
Teleposers

Arcaratus:
-DON'T ACCEPT THIS YET: ONLY FOR VIEWING PURPOSES


src/main/java/WayofTime/bloodmagic/block/BlockTeleposer.java
```diff
@@ -28,6 +30,23 @@ public BlockTeleposer()
     }
 
     @Override
+    public void onNeighborBlockChange(World worldIn, BlockPos pos, IBlockState state, Block neighborBlock)
+    {
+        this.updateState(worldIn, pos);
+    }
+
+    private void updateState(World worldIn, BlockPos pos)
```
WayofTime: Here, you need to cache which state you were in previously and then swap it. So if you were previously unpowered and then became powered, you'd set yourself as powered and then do the operation once. If you are powered previously and are still powered, you do nothing, same with if you were powered and become unpowered. 
Finally, because of this, you have to still do stuff if worldIn.isBlockPowered(pos) is false - this is where if previously powered you set yourself as unpowered.

src/main/java/WayofTime/bloodmagic/client/gui/GuiHandler.java
```diff
@@ -0,0 +1,43 @@
...
+public class GuiHandler implements IGuiHandler
+{
+    @Override
+    public Object getServerGuiElement(int id, EntityPlayer player, World world, int x, int y, int z)
+    {
+        BlockPos pos = new BlockPos(x, y, z);
+
+        switch (id)
+        {
+            case 0:
```
WayofTime: Should make this a constant, so we keep things organized.

WayofTime:
I'll check the rest when you get that fixed - should help with your testing.

PR #536, opened 2016-01-02
Unity textures

BBoldt:
I took out most textures that weren't being used, but still left in a lot of stuff. There is a lot to sort through.

Unity textures are added, a few things needed renamed.

Don't forget that we can still take things from here as we need them: https://github.com/CyanideX/Unity/tree/master/assets/alchemicalwizardry

Oh and I redid the JEI thingy to match.


Yulife:
So this is a new feature exclusive for 1.8? Happy to see our textures in the mod itself :) Only thing missing is omega armour. Not sure if that should be retextured though. :)

TehNut:
The entire armor system is being redone for 1.8. Don't worry about those for now

PR #551, opened 2016-01-08
Updated RitualEffectGrowth to support non-standard farms

grundyboy34:
Updated RitualEffectGrowth's y range from a static value to scaling with the range. This is to add support for mods such as agricraft, in which crops aren't always only on top of soil.


WayofTime:
We are not going to update 1.7.10 any time soon. I'll just merge this for the future use, but please refrain from doing any more 1.7.10 work!

grundyboy34:
Apologies, I had no idea there was a different version on here. Thanks for including the fix though and good work on the mod!

PR #553, opened 2016-01-08
Augmented capacity rune

BrizzleMFD:
changed the In game text for augmented capacity rune for 1.7.10 to the
correct amount each rune actually gives. Book says 1.5k however each actually gives 2k. I know that 1.8.9 each rune gives 1k so i don't want to break the documentation for that. If I did something wrong please don't use my blood for evil.


WayofTime:
We are no longer accepting PRs for 1.7.10. Sorry about this!

BrizzleMFD:
All good.

Blue64:
Could still be merged into the master code for anyone who decides to compile it. A sort of "1.7.10 unsupported" branch where you aren't actually pushing it through, but you'll allow bug-fixes for the code.

TehNut:
I think we're completely done with 1.7.10. Merging and all.

Blue64:
Could someone else take over the 1.7.10 portion then?

PR #562, opened 2016-01-11
Blood Orb fixes, soul gems, and soul snares

BBoldt:
Haven't got around to doing the armors yet, but I figured I would PR Inap's soul gems and soul snare in case you are planning any releases tomorrow.


BBoldt:
Squashed @Yulife's commits into one at the request of @TehNut  ^.^

PR #567, opened 2016-01-14
Fixed up Bound Sword

Arcaratus:
-Fixed up Bound Sword to actually be a bound tool
-Fixed Bound Items charge bar bug


TehNut:
Why doesn't the sword fire the `BoundToolEvent` anymore?

Arcaratus:
It doesn't charge or do anything special

Blue64:
is it planned to eventually? @Arcaratus  @TehNut

Arcaratus:
The Blood Magic IRC channel (#bloodmagic) holds a lot of these content related questions and discussions, you should join to keep up with the latest news and thoughts on upcoming BM content

PR #635, opened 2016-02-19
Update zh_TW.lang

NatsuArashi:


TehNut:
I can only assume this is correct hehe...

PR #639, opened 2016-02-21
Update ru_RU.lang

AngelCry:


WayofTime:
I'm going to assume that these are accurate enough!

AngelCry:
may be i not a best translator, but i do my best :)
where is been possible, i tried to keep old translations, so players, who already played translated BloodMagic can feel itself familiar

PR #641, opened 2016-02-23
Add Capability support to Ritual Sones, Finish TE support as well

Alex-hawks:


TehNut:
I'm trying to understand why this is needed. Way said something about microblocks

Alex-hawks:
The capability portion is needed for me to remake the microblock ritual stones, and I figured that I had to change those classes anyway, so I would finish the Tile Entity support while I was at it.

TehNut:
But _why_ is it required?

Alex-hawks:
because MCMultiPart uses the Capabilities system, instead of using ASM to inject traits like FMP did

PR #685, opened 2016-03-24
Fix ItemBoundSword damage

CJCutrone:
ItemSword has this line:

"this.attackDamage = 3.0F + material.getDamageVsEntity();"

Currently, BloodMagic has this line:

"this.attackDamage = 1.0F + ModItems.boundToolMaterial.getDamageVsEntity();"

This is makes the sword do less damage than it should. (making the pickaxe, axe, and shovel all better in terms of damage.)

Simple change that allows the SuperClass ItemSword handle the attackDamage variable.


TehNut:
Re-submit on 1.9 branch please. 1.8 is no longer being developed.

CJCutrone:
Done. Apparantly you have to make new pull request entirely in order to change branches? So I did that [here](https://github.com/WayofTime/BloodMagic/pull/686).

PR #686, opened 2016-03-24
Fix ItemBoundSword damage [1.9 branch]

CJCutrone:
Redid [this](https://github.com/WayofTime/BloodMagic/pull/685) on the proper branch.


WayofTime:
Probably won't work with the 1.9 branch since it is hardcoded, but whathaveyou.

PR #696, opened 2016-03-31
Fix biomeID selection for the ritual of gaias transformation.

oceth:
The biome id is wrongly assigned an arbitary iteration variable. This works until the loop starts skipping null biomes, where the iteration variable goes out of sync.

This PR changes the selection to the biomeID property of the selected biome object, thus making sure it is always the correct id.


TehNut:
1.7.10 has reached EOL and is no longer being developed.

PR #726, opened 2016-05-06
Update zh_CN.lang

qianzha:


WayofTime:
I'm putting .lang updaters to work, I see! :D

PR #734, opened 2016-05-10
Update zh_CN.lang

qianzha:


qianzha:
Hey Way, I wonder if we can just call the BloodMagic v2.0.0 "BloodMagic2". I'm going to create a new mod page at the mcmod.cn, which is a Chinese wiki of Minecraft mods, because there are too many different between the new version and the old.

qianzha:
Well. I saw the title of the vedios by Direwolf20, which you retweeted. It seems that it's not a problem.

PR #788, opened 2016-06-09
Create zh_CN.lang

qianzha:
The format of GuideAPI seems not nice with Chinese and Chinese punctuation marks. It would happen something like the picture. So I changed something for that in text, like changed where newline characters are, and deleted the "\t" because it seems not work with Chinese. But there is still a "Format error" at entry of snare info2, which is after the recipe (It didn't happen at other pages that translated).

http://i.imgur.com/wms2QQP.png (Chinese with English punctuation marks)
http://i.imgur.com/oeJ6Hgn.png (Chinese with English punctuation marks)
http://i.imgur.com/C1Axqn4.png (Chinese with Chinese punctuation marks, which didn't change \n\t)
http://i.imgur.com/Buc7FWj.png (The "Format error" page that after the recipe)


WayofTime:
When you list % signs you need to do '%%'

qianzha:
So the "%" is where the Format "error" is. 
Then, Is the problem of turning pages with Chinese esay to fix? Although we can just deal with that in text at BM2. This problem happen in 1.7.10, too. And there isn't a newline characters can help.

TehNut:
I can take a look at fixing the overloaded pages. The way it currently works is a character limit (X amount of characters per page before creating a new page). Chinese characters are larger than English so this happens.

On another note, could you not submit untranslated keys in your file? If we change a localization in en_US, untranslated keys in zh_CN will not inherit the changes and will instead display the outdated English localization.

Edit: Also, this entire localization file is currently heavy WIP. By the time it's actually stable, the keys/localizations may be drastically different than they are now. I suggest holding off until we're farther along.

qianzha:
I see. I will wait for that and close this request. Thanks for your work!

PR #818, opened 2016-06-24
Fix Ritual Diviner not placing blocks

Arcaratus:


TehNut:
This includes all the previous commits. You need to use a new branch.

PR #822, opened 2016-06-25
Update zh_CN.lang

qianzha:
Updated and fixed. 
There is a little question about the sanguineBook. Is "Inspectoris" meaning "viewer", and "Scandalum" meaning "stumbling block" which is stumbling to get the next tier?


WayofTime:
It's basically latin for "Block Viewer"

qianzha:
Translated as "Book of Viewer". It seems will be strange that translates word by word.
（Time to sleep in China _(:з」∠)_ ）

qianzha:
Holding Reagent

PR #827, opened 2016-06-28
Bucket recipe fix and Creative Sacrificial Dagger fix

Arcaratus:
-Fixes the Life Essence Bucket recipe
-Make the Creative Sacrificial Dagger creative again!
-Add a config option for the amount of damage done by the Sacrificial Dagger

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/ConfigHandler.java
+++ b/src/main/java/WayofTime/bloodmagic/ConfigHandler.java
@@ -139,6 +139,7 @@ public class ConfigHandler
     // General
     public static int sacrificialPackConversion;
     public static int sacrificialDaggerDamage;
+    public static int sacrificialDaggerConversion;
 
     // Client
     public static boolean alwaysRenderRoutingLines;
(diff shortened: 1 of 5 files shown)


TehNut:
1. I already fixed the bucket recipe. https://github.com/WayofTime/BloodMagic/commit/f500ec70ad9da3076b6f51f17c3472519a2eec13
2. You just broke it again.

Arcaratus:
I did that because someone had an empty bucket as a result so...did you test yours?

TehNut:
I did.

Arcaratus:
Well, I tested the above and it works...

TehNut:
The NBT that you provide there does not carry over to the craft. I moved it to our CraftingHandler. Your commit adds extra unnecessary work.

Arcaratus:
Should I undo it?

WayofTime:
Need to fix the JEI compat then, since it shows an empty bucket.

TehNut:
Will do once I get off work

PR #852, opened 2016-07-15
Update zh_CN.lang

qianzha:


WayofTime:
Oops, forgot this was still here.

PR #862, opened 2016-07-29
[1.7.10]Update all entries of alchemy

AstonIceDog:
Now,I update all entries of alchemy for 1.7.10.But I don't konw whether is it  in time，because you are committing to 1.9.
This part of entries cost me a whole month.Thanks for @qianzha and Hastur.
And other parts of entries are been translated by @qianzha，@Joccob，@MoXiaoFreak and Myth.


TehNut:
1.7.10 is not maintained and hasn't been for _several months_. I don't know why you thought this would be worth the time.

AstonIceDog:
I can only use  imageto show that I'm desperate.

TehNut:
Even if this got pulled, it would never get built or released.

Dream-Master:
I using 1.7.10 so sure why not develop the mod a bit.

TehNut:
Because the more we support 1.7, the longer it lasts which is bad. 

Use a resource pack for this.

PR #938, opened 2016-10-01
More Sigil of the Green Grove fixes and cleanup

josephcsible:
When applying the bonemeal effect, apply it from the player using the sigil,
rather than from a fake player. (Looking at the commit history, this seems
to be this way solely because of overzealous copying and pasting from vanilla
ItemDye.) Also, play the effect for the use of bonemeal, rather than for
breaking a block. Finally, remove a parameter that's no longer necessary.


TehNut:
> Looking at the commit history, this seems to be this way solely because of overzealous copying and pasting from vanilla ItemDye

Sadly, even though I said that nothing should be copied directly from the BM 1.7 codebase, it seems that Arc did that a whole lot.

PR #1008, opened 2016-12-19
Start of translation to spanish

RockGoliath:
There's some ambiguious adjectives. Here's some stuff that the translation could be debated of:

- reagents in spanish doesn't sound all too good to me, I said catalyst instead.
- tinkerer I don't know how to translate, google doesn't either. I said controller instead.
- awakened translates to "despertado" but that doesn't mean quite the same. I left it like that for now, but there's definitely a case to be made for something more akin to imbued or empowered.
- diviner I translated as the tool the guy in the article is using to divine the location of water https://en.wikipedia.org/wiki/Dowsing. "Divinador" could be used too but I personally prefer "zahorí".
- slate. I initially thought "tableta" but after looking at some dictionary translations "loja" I think fits better.
- sigil could go in lots of ways really, I chose the equivalent of the english word sign.
- blood letter's backpack. To bloodlet in spanish is "hacer una sangria". I have no clue what to translate blood letter's backpack to though.

### DIFF
--- a/src/main/resources/assets/bloodmagic/lang/es_ES.lang
+++ b/src/main/resources/assets/bloodmagic/lang/es_ES.lang
@@ -9,7 +9,6 @@ item.BloodMagic.activationCrystal.creative.name=Cristal de Activación Creativo
 
 item.BloodMagic.sacrificialDagger.normal.name=Daga Sacrifical
 item.BloodMagic.sacrificialDagger.creative.name=Daga Sacrifical Creativa
-item.BloodMagic.pack.selfSacrifice.name=Blood Letter's Pack
 item.BloodMagic.pack.sacrifice.name=Escudo de Armas
 item.BloodMagic.daggerOfSacrifice.name=Daga de Sacrificio
 
--- a/src/main/resources/assets/bloodmagic/lang/es_ES.lang
+++ b/src/main/resources/assets/bloodmagic/lang/es_ES.lang
@@ -46,22 +46,6 @@ item.BloodMagic.orb.master.name=Orbe de Sangre de Maestro
 item.BloodMagic.orb.archmage.name=Orbe de Sangre de Archimago
 item.BloodMagic.orb.transcendent.name=Orbe de Sangre Transcendente
 
-item.BloodMagic.reagent.incendium.name=Incendium
-item.BloodMagic.reagent.magicales.name=Magicales
-item.BloodMagic.reagent.sanctus.name=Sanctus
-item.BloodMagic.reagent.aether.name=Éter
-item.BloodMagic.reagent.crepitous.name=Crepitous
-item.BloodMagic.reagent.crystallos.name=Crystallos
-item.BloodMagic.reagent.terrae.name=Terrae
-item.BloodMagic.reagent.aquasalus.name=Aquasalus
-item.BloodMagic.reagent.tennebrae.name=Tennebrae
-item.BloodMagic.reagent.offensa.name=Offensa
-item.BloodMagic.reagent.praesidium.name=Praesidium
-item.BloodMagic.reagent.orbisterrae.name=Orbis Terrae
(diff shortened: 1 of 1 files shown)


WayofTime:
A lot of these assumptions are good. "Sigil" is more of some sort of symbol, as in "I wrote a sigil of power in the ground".

src/main/resources/assets/bloodmagic/lang/es_ES.lang
```diff
@@ -0,0 +1,164 @@
...
+item.BloodMagic.sigil.haste.name=Signo de Velocidad
+item.BloodMagic.sigil.suppression.name=Signo de Supresión
+item.BloodMagic.sigil.magnetism.name=Signo de Magnetismo
+item.BloodMagic.sigil.fastMiner.name=Signo del Minero Veloz
+item.BloodMagic.sigil.seer.name=Signo del Vidente
+item.BloodMagic.sigil.phantomBridge.name=Signo del Puente Fantasmal
+item.BloodMagic.sigil.whirlwind.name=Signo del Remolino
+item.BloodMagic.sigil.enderSeverance.name=Signo de División Ender
+item.BloodMagic.sigil.holding.name=Signo de Almacenamiento
+item.BloodMagic.sigil.holding.display=&r%s: &o&n%s
```
TehNut: This line seems to be the same as the English translation

src/main/resources/assets/bloodmagic/lang/es_ES.lang
```diff
@@ -0,0 +1,164 @@
...
+item.BloodMagic.slate.ethereal.name=Loja Etérea
+
+item.BloodMagic.orb.weak.name=Orbe de Sangre Débil
+item.BloodMagic.orb.apprentice.name=Orbe de Sangre de Aprendiz
+item.BloodMagic.orb.magician.name=Orbe de Sangre de Mago
+item.BloodMagic.orb.master.name=Orbe de Sangre de Maestro
+item.BloodMagic.orb.archmage.name=Orbe de Sangre de Archimago
+item.BloodMagic.orb.transcendent.name=Orbe de Sangre Transcendente
+
+item.BloodMagic.reagent.incendium.name=Incendium
```
TehNut: Since these are Latin names, they shouldn't get translated. They also don't need entribs in here

TehNut (changes requested):

src/main/resources/assets/bloodmagic/lang/es_ES.lang
RockGoliath: I didn't touch any but aether, since that one has a proper translation in spanish (éter). Makes sense though, I'll remove the stuff that is not changed that I left in the pull request by mistake.

src/main/resources/assets/bloodmagic/lang/es_ES.lang
RockGoliath: Yeah, I was told not to touch stuff that had weird symbols in it. I should have just deleted it.

TehNut: The single characters prefixed with a % (only s in this case) are for string replacements. The ones prefixed by & are for Minecraft formatting codes for colors and such.

TehNut (approved):

WayofTime:
Raw means impure, like uncooked food is raw.
Yes, tiled refers to tiles.
Blood Letter isn't a profession, just describing the fact you are letting blood when you use this.
Draft refers to a drink, like a draft of beer.
The "Small" refers to the size of the stone bricks in the wall.
Swiffer is a type of broom that picks up dust, it's a brand-name thing.
You can take the movement meaning for this joke if it isn't translatable.
Sigil is old-fashioned in english as well, so it may be best to use the old fashioned translation. If that doesn't work, go for symbol.

Issue #1036, opened 2017-02-04
Living armor self sacrifice update not working


pksiazek:
#### Issue Description:
No matter how many self sacrifices I make living armor is not getting Tough Palms upgrade.

#### Steps to reproduce:
1. Craft Living Armor set (with available upgrade points)
2. Self sacrifice on blood altar

#### Affected Versions
- BloodMagic: 1.9.4-2.1.0-69
- Minecraft: 1.10.2
- Forge: 12.18.3.2221

#### Technical details
ItemSacrificialDagger.onItemRightClick() seems to have a lot of obsolete code (all sacrifical logic is done by onPlayerStopperUsing() method).
In normal flow execution can't get past line 110 so SacrificeKnieUsedEvent is never called.

pksiazek:
https://github.com/WayofTime/BloodMagic/pull/1037

I believe this should fix issue :D

TehNut:
Oh, this should be fixed and released as of the updates yesterday.

PR #1037, opened 2017-02-04
#1036 Fixed Living Armor sacrifical upgrade

pksiazek:


TehNut:
How does removing all the sacrifice code from the dagger fix the Living Armor

pksiazek:
Because this code is unused anyway as I explained in issue.
Current working code is in PlayerSacrificeHelper

WayofTime:
I tested the changes and this makes the dagger not work when you do not have incense.

EDIT: I think only the changes in PlacerSacrificeHelper are needed.

pksiazek:
Damn.
You are completely right.
I fixed PR.

PR #1052, opened 2017-02-14
Update zh_CN.lang

3TUSK:
~~Don't merge this pull request yet, as this pr contains large amount of text, and I cannot 100% guarantee there is nothing wrong. I can only squash commits *only if* I am sure it is ready to go.~~

Edit: Now ready to go!

Summary:
* Use full-width punctuation in `assets/bloodmagic/lang/zh_CN.lang`.
* A few items got better names.
* A few sentences were re-phrased so that they sound better.
* Full translation of `assets/bloodmagicguide/lang/zh_CN.lang`.
  * With effort of explaining all possible references and latin phrases. I didn't include explanations because the original context didnt do so. Still struggling with remembering latin vocabulary, I may need save some time for self-studying latin -_- 

Summoning @qianzha @TartaricAcid for double check.

~~Todo:~~ No to-do left now.
* ~~Translation of "Demon(ic) Will" haven't settled down yet.~~
* ~~"Architect"~~
* ~~"Vengeful"~~


TehNut:
Don't worry about squashing commits, I can do that on my end directly through Github

3TUSK:
Ah i forget squash merge. I was just reading contributing.md....

TehNut:
Yeah... That's out of date... The CONTRIBUTING is the least of my worries right now >.>

3TUSK:
Just confirmed, the Chinese characters will still cause overflow as shown below (Ritual Master, Introduction):
image

Could adding more `\n\t` solve this problem temporarily?

3TUSK:
~~@TehNut Found the issue - Apache lang3's `WordUtils.wrap` seems to have no effect when `wrapLongWords` is false and there is no space in target string (this is the case in Chinese, as Chinese does not use space to break words).~~

~~Easiest solution would be adding spaces as *page break*. Commit incoming...~~

~~Edit: further debug indicates that translated text cannot fit into wrap length of 370... guess I need a robust fix...~~

Edit *2: Never mind, after debugging for couple hours I realized that I could use some black magic to get it work. Commit is on the way.

3TUSK:
~~Now just wait for @qianzha and @TartaricAcid~~ ...

~~Edit: get approval from @TartaricAcid~~

Edit: ready.

TartaricAcid (approved):

qianzha (approved):

3TUSK:
@TehNut ping

TehNut:
All good to go?

3TUSK:
Yes

Issue #1065, opened 2017-02-25
1.11.2 entity.sentient specter.name Killing everyone


maxsil:
#### Issue Description:
Dropping a sentient sword to a friend casused entity.sentient specter.name to spawn, it's invisible, i can't see it's hitbox with f3+b restarting server does nothing.  It's killing everyone on the server if they're in the same chunk or whatever as me, if they respawn and come back they die, they don't need to attack me. 

I've tried using the kill command but it doesn't work, i don't know the entities name (and variations of entity.sentient specter dont' work)
#### Edit:

Oh yeah i solved it by running a /kill @e, I even got my sword back. So if anyone else is being haunted, suicide is the only way out ;)
#### Steps to reproduce:

1. Craft sentient sword
2. have more than 1000 will
3. drop sword
...

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: 2.1.4-73
- Minecraft: 1.11.2
- Forge: 13.20.0.2228

TehNut:
Oh dear. I'm sorry but that is hilarious.

TehNut:
Bigger issue than previously thought. None of the entities render. 

E: still hilarious

maxsil:
We didn't know what was going on at first, i was just giving my buddy a sentient sword and then people around me started dying, we resstarted the server and everything. Then i made a joke about the house being haunted. Which it turns out it literally was

TehNut:
In my testing, `/kill @e[type=!Player]` was correctly killing the specters, btw. So worst case, just run that. It'll kill all item frames though. If you're familiar with Vanilla commands you could probably find a way around that.

Gaelan:
/kill @e[type=bloodmagic:scepter]?

edit: sorry about the ping

maxsil:
Hey thanks for the replies! Yeah we solved it by killing everything ( which isn't that big of a deal thanks to the EXP book in this mod :) ), but we're bound to accidentally summon it again, and we'll run those commands then!

NobleWyrm:
Commenting here as I work on fixing this since I'm quite rusty on Forge specific code. I've still got my general debugging skills though, just posting things here as I find them in case someone else sees the fix before I figure it out from what I find.

[...]

PR #1087, opened 2017-03-07
Entity Registration Fix

NobleWyrm:
Replaced the Entity registration code. 

Before, the code was manually adding entities to the GameRegistry. Doing this prevents the engine from fully registering the added entities in all the proper places. The result was that the Client was never informed of when these entities were created or what they were doing, as all of that behavior is managed by the EntityRegistry.

Changing to the proper EntityRegistry calls fixes issue #1065.

It should be noted that I wasn't super sure what to enter as the values for the  tracking components for the entities, so I set all of them to send tracking updates to the player as long as they are within  4 chunks, send those tracking updates every 3 ticks, and to include velocity with the tracking data. If this should be changed later, it can be done in another pull request.

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/registry/ModEntities.java
+++ b/src/main/java/WayofTime/bloodmagic/registry/ModEntities.java
@@ -1,6 +1,8 @@
 package WayofTime.bloodmagic.registry;
 
-import net.minecraftforge.fml.common.registry.EntityEntry;
+import WayofTime.bloodmagic.BloodMagic;
+import WayofTime.bloodmagic.api.Constants;
+import net.minecraft.util.ResourceLocation;
 import WayofTime.bloodmagic.entity.mob.EntityCorruptedChicken;
 import WayofTime.bloodmagic.entity.mob.EntityCorruptedSheep;
 import WayofTime.bloodmagic.entity.mob.EntityCorruptedSpider;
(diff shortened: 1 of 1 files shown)


src/main/java/WayofTime/bloodmagic/registry/ModEntities.java
```diff
@@ -11,21 +13,22 @@
...
-        GameRegistry.register(new EntityEntry(EntitySentientArrow.class, "SoulArrow").setRegistryName("SoulArrow"));
-        GameRegistry.register(new EntityEntry(EntityMeteor.class, "Meteor").setRegistryName("Meteor"));
-        GameRegistry.register(new EntityEntry(EntitySentientSpecter.class, "SentientSpecter").setRegistryName("SentientSpecter"));
-        GameRegistry.register(new EntityEntry(EntityMimic.class, "Mimic").setRegistryName("Mimic"));
-        GameRegistry.register(new EntityEntry(EntityCorruptedZombie.class, "CorruptedZombie").setRegistryName("CorruptedZombie"));
-        GameRegistry.register(new EntityEntry(EntityCorruptedSheep.class, "CorruptedSheep").setRegistryName("CorruptedSheep"));
-        GameRegistry.register(new EntityEntry(EntityCorruptedChicken.class, "CorruptedChicken").setRegistryName("CorruptedChicken"));
-        GameRegistry.register(new EntityEntry(EntityCorruptedSpider.class, "CorruptedSpider").setRegistryName("CorruptedSpider"));
+        int entities = 0;
+        EntityRegistry.registerModEntity(new ResourceLocation(Constants.Mod.MODID+":BloodLight"),EntityBloodLight.class, "BloodLight",++entities, BloodMagic.instance,16*4, 3, true);
```
TehNut: Missing some whitespacing in these. Also, use the the RL constructor with 2 string inputs.

TehNut (changes requested):

NobleWyrm:
I have replaced the ResourceLocation calls with the two-argument version and added more consistent white-space.

TehNut (approved):

TehNut:
Just curious, is there any reason you're using `16*4` for each of these instead of 64?

NobleWyrm:
Seems a bit more clear than just 64, since it makes it easier to understand that it is 4 chunks at a glance

TehNut:
Good enough of a reason for me ¯\\\_(ツ)_/¯

PR #1164, opened 2017-06-22
Add crop support for Pam's Harvestcraft and Roots herbs

gchpaco:
This adds support for Pam's many crops and Roots herbs to the Harvest Moon ritual.  Rather than enumerate all the Pam's crops, it uses reflection to get a list from Harvestcraft itself.

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/ritual/harvest/HarvestHandlerPlantable.java
+++ b/src/main/java/WayofTime/bloodmagic/ritual/harvest/HarvestHandlerPlantable.java
@@ -1,5 +1,9 @@
 package WayofTime.bloodmagic.ritual.harvest;
 
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.InvocationTargetException;
+
 import WayofTime.bloodmagic.api.BlockStack;
 import WayofTime.bloodmagic.api.iface.IHarvestHandler;
 import WayofTime.bloodmagic.api.registry.HarvestRegistry;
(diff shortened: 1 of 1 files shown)


src/main/java/WayofTime/bloodmagic/ritual/harvest/HarvestHandlerPlantable.java
```diff
@@ -98,4 +113,37 @@ private static void addThirdPartyCrop(String modid, String regName, int matureMe
...
+            String className = "com.pam.harvestcraft.blocks.CropRegistry";
+            Class<?> registry = ReflectionHelper.getClass(loader, className);
+            Field names = ReflectionHelper.findField(registry, "cropNames");
+            Method getCrop = registry.getMethod("getCrop", String.class);
+            for (String name : (String[])names.get(null)) {
+                Block crop = (Block) getCrop.invoke(null, name);
+                HarvestRegistry.registerStandardCrop(crop, 3);
+            }
+        } catch (NoSuchMethodException e) {
+            FMLLog.warning("HarvestCraft integration cancelled; unable to find crop name mapper");
```
TehNut: I would rather these used `BloodMagic.instance.getLogger()`

TehNut (changes requested):

TehNut:
Do all Pam's crops really mature at an age of 3? I thought they went up to 7.

Since you're already reflecting all the things and Pam uses BlockCrops, you could probably cast to that and call `getMaxAge()`.

gchpaco:
I've used the new logger, and called `getMaxAge` per your thoughts.  The JavaDoc situation for Forge 1.11 is not the best, and I didn't actually realize `getMaxAge` existed.

Pam crops do actually mature at 3 in the version of the source code we have; but because that version is not in any way current it might have changed since then.  I suppose I could have decompiled it, but this project didn't seem merit that kind of detail.

TehNut:
>Pam crops do actually mature at 3 in the version of the source code we have

Huh, didn't know that. I don't use Pam's so...

TehNut (approved):

TehNut:
Thanks!

PR #1189, opened 2017-09-29
Fix operator precedence causing wrong items to be crafted in Rhythm of the Beating Anvil

calmofthestorm:
Missing parentheses cause incorrect items to be crafted when multiple recipes are used in the ritual, since the check that items are equivalent will return true for all undamaged items. If there is a valid crafting recipe for the same shape and the block chosen, it will be crafted, otherwise, the inputs will simply disappear. This basically means that crafting more than one recipe in the same recipe is unlikely to ever work reliably. This should not apply to crafting only a single recipe.

I understand that 1.7.10 is EOL so this will never be fixed, but I wanted to document this problem so that it doesn't come along when/if this ritual is ported to newer versions, and to create a record for others googling the Rhythm of the Beating Anvil for crafting the wrong items, losing items, etc in 1.7.10.


TehNut:
Closing since 1.7 is unsupported.

PR #1193, opened 2017-11-02
Update to 1.12.2

Spencer-H:
Update Forge to the recommended version for 1.12.2

Update the MCP mapping to the most recent.

Update Hwaila to the newest version.

Replace functions that have been renamed.
No functionality was changed here beyond a few functions being merged.
IE: getDistanceSqToEntity is now just getDistanceSq as it now makes use
of specialization instead of having multiple functions.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,7 +39,7 @@ dependencies {
 }
 
 minecraft {
-    version = "${mc_version}-${forge_version}"
+    version = "${mc_version}.2-${forge_version}"
     runDir = "run"
 
     replace "@VERSION@", project.version
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,11 +2,11 @@ mod_name=BloodMagic
 package_group=com.wayoftime.bloodmagic
 mod_version=2.2.0
 mc_version=1.12
-forge_version=14.21.1.2443
+forge_version=14.23.0.2491
 curse_id=224791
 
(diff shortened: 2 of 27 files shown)


TehNut:
Unless I absolutely need a feature from 1.12.2, I am not going to update to it. Mapping and dependency changes should only be changed by project managers.

Sunconure11:
@TehNut Any reason for the delay in builds for later versions from your end? I'm aware Way is busy. I managed to port it, saw this shot down commit, and now I'm curious.

Sorry about the necromancy.

TehNut:
1. Lack of motivation.
2. Rewriting a good chunk of the mod.
3. I've picked up a few projects outside of Minecraft that are eating a lot of my time.

Also, this PR did absolutely nothing but update dependencies and mappings. It was literally useless.

It also did it in an extremely stupid way. Like, really?

Sunconure11:
It seemed to be somewhat functional when I tried. Though some recipes broke.

Granted, so far, my port is just a mappings update, like his, but I'll see to fixing some of the recipe bugs I saw, if you ever want to look it over.

screenshot 2017-12-15 14 41 53

Haste reagent recipe, for instance. Actually, all of the reagent recipes are broken in some form or another.

TehNut:
I would rather you dump whatever you have and work from where the main branch is at. I wont be pulling anything that changes versions.

PR #1244, opened 2018-03-08
This *should* in theory work

Arcaratus:
I think I did the PR right. Right...?

Should fix the mess of the charge bar -> new "charging" textures are needed for bound tools.
Nukes BlockStack and ItemStackWrapper usage within bound tools.


TehNut:
Stop using the damage bar. Use the property overrides you created in the model like I told you to.

Arcaratus:
My bad, forgot to remove those methods

PR #1250, opened 2018-03-14
EnumDyeColor.getColorValue() Fix

FaeyUmbrea:
Why I made this:

`EnumDyeColor.getColorValue()` has become a client-side only method. This means that using any dyable recipie in SMP will instantly crash the server and leave the world unable to load due to the dye being in the crafting bench still. 

What I changed:

Basically I made a clone of the `EnumDyeColor` class and removed the client side only restrictions. The class is basically the same compared to the minecraft sources and serves the same functionality as the original class with the addition of a static method to construct the `EnumServerDyeColor` object from stack metadata, the method works exactly the same as the `getBaseColor()` method on the `ItemBanner` class. The classes can not be used interchangably since I didn't factor in a way to convert between `EnumDyeColor` and `EnumServerDyeColor`. Though looking at the codebase, there are only two uses for the original Enum class and they are both covered by my update.

The class is the best solution for the issue I could come up with while trying to fix this all night. Also, I did work on the current releases commit in order to compile a version that works on my server until an official update fixing the issue is uploaded and then merged the 1.12 branch into my fix branch in order to be up to date with the bases 1.12 branch.


FaeyUmbrea:
Also this is the crash report the change is based on. I've confirmed that this is indeed said method, but see for yourself if you want to.

[crash-2018-03-14_03.39.32-server.txt](https://github.com/WayofTime/BloodMagic/files/1809764/crash-2018-03-14_03.39.32-server.txt)

TehNut:
Dear god no this is terrible

FaeyUmbrea:
Well, I'm not entirely happy with how I solved it, but lacking any shape or form of proper documentation available to me thats pretty much the best I could come up with. I haven't had much experience with minecraft mods and I think you could've been a bit more explanatory as to what the issue is. 

Also besides that, I'm unsure if your fix is going to work. I imagine there would be a reason for the method to be only available on the client side. My best guess would be that the values are just not implemented on serverside, which would, when reading them out like you wrote it, result in all dyes ending up being the same color. 

I did a build including the latest commit and at least the server doesn't crash anymore. But I have no idea which recipie caused the crash in the first place since one of my players did report it to me. 

Then again, I could be overthinking the reasoning behind making the method client only, I have no idea, but it seems just stupid to me to make a value client only without any good reason behind it.

TehNut:
If you could stop assuming how the internals of the game works, that'd be great. I've been working with this codebase for over 3 years now. I know how `@SideOnly` works. If you'll note, it is *not on the field I am accessing*. The field exists. The values exist. The getter method does not.

>you could've been a bit more explanatory as to what the issue is.

You literally copy-pasted an entire Vanilla class (Even all the irrelevant data. Wonderful!) because you wanted a single value. What *isn't* the issue here?

> it seems just stupid to me to make a value client only without any good reason behind it.

Mojang's obfuscator checks what sides methods are being used on and only adds them to the jars that require them.

FaeyUmbrea:
I don't get why you're taking this as a personal insult, but maybe thats due to english not being my first language and me misjudging my or your tone. 

I'm absolutely not used to working with minecrafts codebase, I'm a server admin whose server crashed because my player crafted something and this is the best thing I could come up with at what was at the time 6 am. But on top of that,

My understanding was, knowing that minecraft was obfuscated, that the side-only annotation was introduced by the mod framework in order to seperate client mod code from server mod code to not try loading rendering classes on the server end, which was a long term issue the mod framework had. 

I'm a bit hiding behind facts and I do agree that it wasn't my best idea to post the pull request, my apologies for that. At least I learned something about minecrafts very confusing codebase. 

I do have a few furhter questions, but I don't want to litter the repository with them, I did add you on steam for that purpose. But I'll go get some sleep first.

kingpotato2000:
I think the thing to have done here is just leave a bug report, rather than try do something you are not completely familiar with

FaeyUmbrea:
Yeah, I'm well aware that that was stupid, sadly only after doing it.

kingpotato2000:
Well, at least it's sorted now

kingpotato2000:
Or soon anywau

kingpotato2000:
*anyway

FaeyUmbrea:
Yeah, it seems though, that TehNut is a bit fed up with me. I'd love to ask him a few questions though, since I do want to be able to collaborate and write mods in my off time.

kingpotato2000:
Yes, but it's understandable why when someone has worked on something for a long time then someone else tries something that doesn't really work with everything else

FaeyUmbrea:
Well, guess I shouldn't try this kind of stuff anymore, was worth a try

kingpotato2000:
If you not sure, leave a bug report

PR #1297, opened 2018-05-04
Require water potion instead of uncraftable potion

starlightsys:
Fixes issue where cutting fluid is uncraftable.


TehNut:
Huh. I fixed it further down but not here

PR #1337, opened 2018-06-09
Updated de_DE.lang

PiscesdanAT:
Someone mixed up dusk and dawn


TehNut:
Google Translate checks out 👍

PR #1347, opened 2018-06-21
Fix the Blood Tank bounding box

Arcaratus:
-Made the Blood Tank less thicc
-with pics

# **BEFORE**

# **AFTER**


TehNut:
You had me at less thicc.

PR #1380, opened 2018-07-29
Fixed a null pointer exception with certain modded blocks with the Ritual of Magnetism

ChrisThirtle:
After looking through some of the choices I could make with rewriting this little bit of code, I decided to keep it simple and keep the use of the getPickBlock() method.
I made it so the Ritual of Magnetism passes a RayTraceResult from the MRS to the target block when calling the getPickBlock() method. This stops it from throwing NPEs for any kind of multipart block that bothers to override getPickBlock() to let you select the actual multipart being looked at.
It's rather unlikely that an ore would ever be a multipart block, so the fact that the RayTraceResult is just made up largely out of thin air isn't really relevant to its job. And in the case that an ore is a multipart that cares about the exact RayTraceResult, the previous code would have just crashed the server, so it wouldn't have found it anyway.

When the mod updates to Minecraft 1.13 all of this can be tossed out in favor of the new vanilla interface for getting item information from blocks, but for now, this fixes the crash.

The Ritual still passes a null value for the EntityPlayer argument of getPickBlock(), but I didn't quite feel confident in setting up a FakePlayer just for the ritual.
Especially since I have no idea how to do that in anything resembling the correct way.

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/ritual/types/RitualMagnetic.java
+++ b/src/main/java/WayofTime/bloodmagic/ritual/types/RitualMagnetic.java
@@ -12,6 +12,8 @@ import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.RayTraceResult;
 import net.minecraft.world.World;
 import net.minecraftforge.oredict.OreDictionary;
+import net.minecraftforge.common.util.FakePlayer;
+import net.minecraftforge.common.util.FakePlayerFactory;
 
 
 import java.util.function.Consumer;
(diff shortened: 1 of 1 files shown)


src/main/java/WayofTime/bloodmagic/ritual/types/RitualMagnetic.java
```diff
@@ -68,8 +72,10 @@ public void performRitual(IMasterRitualStone masterRitualStone) {
                 while (i <= radius) {
                     while (k <= radius) {
                         BlockPos newPos = pos.add(i, j, k);
+                        Vec3d newPosVector = new Vec3d(newPos);
                         IBlockState state = world.getBlockState(newPos);
-                        ItemStack checkStack = state.getBlock().getPickBlock(state, null, world, newPos, null);
+                        RayTraceResult fakeRayTrace = world.rayTraceBlocks(MRSpos, newPosVector, false);
+                        ItemStack checkStack = state.getBlock().getPickBlock(state, fakeRayTrace, world, newPos, null);
```
TehNut: Rather than passing a null player, look at how the Ritual of Crushing handles a fake player

https://github.com/WayofTime/BloodMagic/blob/b441e7fc1e0ce7587d70ae140c83681933994371/src/main/java/WayofTime/bloodmagic/ritual/types/RitualCrushing.java#L290-L292

TehNut (changes requested):

TehNut (approved):

PR #1387, opened 2018-08-03
Ritual portal delay / Bound Pickaxe fix/improvement

### LINKED ISSUE #1001: Gate of the Fold graphical bug and usability problem
#### Issue Description:
1. Graphical: When using the Bound Pickaxe's right click effect on a Gate of the fold ritual the MRS will break and the ritual will cease to function but the portal visual effect will remain. http://prntscr.com/dhuoce
2. Usability: The portal will teleport you after a few seconds if you remain standing inside of it, even after teleporting once. This means that in an environment with any lag at all you're trapped in a looping teleportation until you luck out and manage to get out of the way.

#### What happens:
2. Trapped in a teleportation loop

2. Not get trapped in a teleportation loop

#### Steps to reproduce:
1. Break the MRS with a bound Pickaxe's right click effect
2. Use the Gate of the fold ritual

____
- BloodMagic: BloodMagic-1.9.4-2.1.0-68.jar
- Minecraft: 1.10.2
- Forge:  forge-12.18.2.2171

keraldi:
RitualPortal (Gate of the Fold) now works similar to the Nether Portal in respect of teleportation cooldown (you have to leave the portal first). This also affects teleposers (if a teleportation within 10 ticks of the previous teleportation is attempted, nothing happens).

Bound Pickaxe now destroys blocks properly with removedByPlayer(), this triggers a block update and everything that happens if a player destroys a block in creative mode (Drops are put directly into the players inventory through a function that is executed before the block removal).

this closes #1001


src/main/java/WayofTime/bloodmagic/item/ItemBoundTool.java
```diff
@@ -250,4 +250,24 @@ protected static void dropStacks(Multiset<ItemStackWrapper> drops, World world,
                 world.spawnEntity(new EntityItem(world, posToDrop.getX(), posToDrop.getY(), posToDrop.getZ(), stack.toStack(count)));
         }
     }
+
+    protected void sharedHarvest(ItemStack stack, World world, EntityPlayer player, BlockPos blockPos, BlockStack blockStack, HashMultiset drops){
+        boolean silkTouch = EnchantmentHelper.getEnchantmentLevel(Enchantments.SILK_TOUCH, stack) > 0;
```
TehNut: Silk Touch and Fortune levels should be passed in as arguments. The levels won't be changed between each block check.

TehNut (changes requested):

keraldi:
I should have known better. For some reason I thought having less imports and not passing it is better, completly ignoring that it would be inside of 3 for-loops for every tool.

TehNut (approved):

Issue #1305, opened 2018-05-13
Well of Suffering lag buildup


ouroborus:
#### Issue Description:
The well of suffering increases server lag over time if not allowed to "rest".

I have a combination of a L5 blood altar, well of suffering, and DE spawner. The spawner is fast enough that the well is never without something to kill. Blood is siphoned off if the altar is near full. There is an orb in the alter to keep the well powered. There are enough upgrades to keep things from becoming bottlenecked.

Leaving this setup active for a few hours is enough to increase TPS lag to the point that the server is running at full capacity. (This is not an immediate effect; it builds up over time. I have a pretty beefy server: I run ATM3 and typically have utilization of about 60% of a single threading core.) Turning off the spawner for a few minutes relieves the lag.
____
#### Affected Versions:

- BloodMagic: 2.2.8-91
- Minecraft: 1.12.2
- Forge: 14.23.33.2669

TehNut:
I'll stick a profiler on it tonight, but I do not think this is a BM issue.

ouroborus:
Alright. Be sure the spawners you use are able to keep up with the well. There needs to always be some creatures available. If the well is given a break, the lag goes away.

TehNut:
I just ran it for 2 hours and the server was at a constant 20TPS. The WoS was being fed by a Tier 5 Soul Shard so it had a constant supply of 30 entities to draw from at all times. After checking with Lag Goggles, The altar was taking 34μs/t and the MRS was taking 23μs/t. The highest timing was actually a block of lava at 337μs/t. 

This was performed in a minimal environment made up of:

* forge-1.12.2-14.23.3.2689-universal.jar
* BloodMagic-1.12.2-2.2.12-97.jar
* Guide-API-1.12-2.1.5-60.jar
* LagGoggles-FORGE-1.12.2-4.1.jar
* SoulShardsRespawn-1.12.2-1.0.4-6.jar

ouroborus:
I'll give laggoggles a shot on my build and let you know.

ouroborus:
How were you keeping the blood alter from filling up?

TehNut:
Tier 4 and 5 were filled with augmented capacity runes. The altar was filling a Transcendent Blood Orb.

TehNut:
After another 2 hours, there is no change. I am almost certain this is not an issue with the WoS.

ouroborus:
[...]

Issue #998, opened 2016-12-06
Suggestion - Prevent well of suffering from dropping items


Castone22:
It might be nice if it were possible to apply a reagent or use the tinkerer to stop well of suffering from dropping items and or xp.  It's really common for players to just set it up and let it litter the ground with stuff while they use another spawner and grinder system for items.

VT-14:
The people who are too lazy to set up a Call of the Zephyr ritual to collect the free mob drops are still going to be too lazy to feed the ritual demon will to remove them completely.  In order for this change to be effective, you would have to make no drops the default.  I see that as both good for lag prevention, and bad for punishing people who are collecting and using those mob drops because of a few irresponsible players.

Whatever direction is taken, it should probably also be applied to the Demon Will Crystal growing ritual (whose name I have yet to memorize).

Do the rituals drop xp?  I just assumed they didn't.  Perhaps a will upgrade could be having rituals drop player kill items, and maybe looting as another one?

TehNut:
I don't believe they drop experience or player specific drops (like Blaze Rods)

bookerthegeek:
I would prefer, from a server standpoint, if you defaulted to no mob drops, and an additional  something to have the mobs drop loot.

This way you could have different types of things you could add,  to change from noisily loot to only xp to player drops .

KingEdward9830:
I remember playing on a server and I forgot to put in something to collect the items, nearly crashed the entire server... I agree with bookerthegeek on this one.

TehNut:
I don't think we'll disable drops by default, but if we add an augment to remove them, I'm sure we can have a config to toggle the how it performs.

keraldi:
Config option exists in newest versions.

PR #1388, opened 2018-08-03
Well of Suffering item drop control

keraldi:
- Added a new DamageSource for the Well of Suffering
- Added an event handler for death through Well of Suffering
- Added a config option to enable (true) or disable (false) Well of Suffering Mob drops.

part of #998, #1305

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/ritual/RitualManager.java
+++ b/src/main/java/WayofTime/bloodmagic/ritual/RitualManager.java
@@ -10,9 +10,13 @@ import net.minecraft.util.DamageSource;
 import net.minecraftforge.common.config.Configuration;
 import net.minecraftforge.fml.common.discovery.ASMDataTable;
 
-import java.util.*;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 public class RitualManager {
+    public static final DamageSource RITUAL_DAMAGE = new DamageSource("ritual_damage").setDamageBypassesArmor();
 
     private final Map<String, Ritual> rituals;
     private final Map<Ritual, String> ritualsReverse;
@@ -20,7 +24,7 @@ public class RitualManager {
     private final Map<String, ImperfectRitual> imperfectRituals;
     private final Map<ImperfectRitual, String> imperfectRitualsReverse;
     private final Configuration config;
-    public static final DamageSource RITUAL_DAMAGE = new DamageSource("ritual_damage").setDamageBypassesArmor().setDamageAllowedInCreativeMode();
+
 
     public RitualManager(Configuration config) {
         this.rituals = Maps.newTreeMap();
--- a/src/main/java/WayofTime/bloodmagic/ritual/RitualManager.java
+++ b/src/main/java/WayofTime/bloodmagic/ritual/RitualManager.java
@@ -6,6 +6,7 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import net.minecraft.block.state.IBlockState;
+import net.minecraft.util.DamageSource;
 import net.minecraftforge.common.config.Configuration;
 import net.minecraftforge.fml.common.discovery.ASMDataTable;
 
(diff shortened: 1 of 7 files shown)


TehNut:
Pretty sure the damage source needs a lang entry for death mesages

keraldi:
I'll look into it. But since it cannot damage players, it should be rarely needed.

TehNut:
Named entities have death messages

keraldi:
I actually named some of my pets that I put below the ritual - no death message, at least not in SSP. I'll just add the string to the blood magic language file, I guess...?

keraldi:
Completed.

src/main/java/WayofTime/bloodmagic/ritual/RitualManager.java
```diff
@@ -19,6 +20,7 @@
     private final Map<String, ImperfectRitual> imperfectRituals;
     private final Map<ImperfectRitual, String> imperfectRitualsReverse;
     private final Configuration config;
+    public static final DamageSource RITUAL_DAMAGE = new DamageSource("ritual_damage").setDamageBypassesArmor().setDamageAllowedInCreativeMode();
```
TehNut: Constants belong at the top of a class, not below the member fields.

src/main/java/WayofTime/bloodmagic/ritual/types/RitualWellOfSuffering.java
```diff
@@ -17,6 +16,8 @@
 import java.util.List;
 import java.util.function.Consumer;
 
+import static WayofTime.bloodmagic.ritual.RitualManager.RITUAL_DAMAGE;
```
TehNut: Single use static imports are 🤢

src/main/java/WayofTime/bloodmagic/ritual/types/RitualForsakenSoul.java
```diff
@@ -20,6 +19,8 @@
 import java.util.List;
 import java.util.function.Consumer;
 
+import static WayofTime.bloodmagic.ritual.RitualManager.RITUAL_DAMAGE;
```
TehNut: Single use static imports are 🤢

src/main/java/WayofTime/bloodmagic/util/handler/event/GenericHandler.java
```diff
@@ -81,6 +80,8 @@
 
 import java.util.*;
 
+import static WayofTime.bloodmagic.ritual.RitualManager.RITUAL_DAMAGE;
```
TehNut: Single use static imports are 🤢

TehNut (changes requested):

keraldi:
*Using stuff suggested by the IDE is not always cleanest* 
(for the stuff in package ritual, it's just RitualManager.RITUAL_DAMAGE)
GenericHandler now has inline WayofTime.bloodmagic.ritual.RitualManager.RITUAL_DAMAGE

Renamed the LivingDropsEvent to onRitualDeath

Something is sorting the imports in alphabetical order... I need to figure that out.

src/main/java/WayofTime/bloodmagic/util/handler/event/GenericHandler.java
```diff
@@ -405,6 +410,15 @@ public static void onLivingDrops(LivingDropsEvent event) {
         }
     }
 
+    @SubscribeEvent
+    public static void onRitualDeath(LivingDropsEvent event){
+        if(!ConfigHandler.values.wellOfSufferingDrops){
+            if(event.getSource().equals(WayofTime.bloodmagic.ritual.RitualManager.RITUAL_DAMAGE)) {
```
TehNut: Why fully qualified class name?

TehNut (changes requested):

src/main/java/WayofTime/bloodmagic/util/handler/event/GenericHandler.java
keraldi: Because you were like "static import bah" and I was like "okay, so I'll not go for an import at all because it's only needed once" and went for the fully qualified class name.

And at that particular place, it only works if fully qualified.

TehNut (approved):

PR #1389, opened 2018-08-04
Third stage of "Dagger of Sacrifice"-glow

### LINKED ISSUE #1034: [Suggestion] Third texture for Sacrificial Dagger
#### Suggestion Description:
Oftentimes it's slightly difficult to see when the Incense Bonus is full, and thus I suggest changing the dagger's texture the third time to signify, that it doesn't receive any more bonus.

#### Subject of matter:
It's difficult to know if Incense Bonus is full without looking at the Incense Altar, which can prove troublesome for metagaming - for instance, placing an Incense Altar beneath the Blood Altar

#### Suggestion of improvement:

Add a third texture for Sacrificial Dagger, triggered when the Incense Altar's max bonus is reached. This can be done by saving the altar's max bonus whenever the player walks by one.

As an alternative, use the third texture when the dagger is charging, switching it to the normal "charged" dagger (or vice versa) when it is not charging, but has some bonus. - this can be done more easily, I assume.

#### Example of implementation:
https://puu.sh/tKkTl.png

#### Possible addition:
Separately texturing the dagger for 20%, 120%, 200% and 300% bonus, this would however require a willing artist

keraldi:
Added Soul Fray check.
Sacrificial dagger now glows even more if you're fully prepared!
Added NBT field for maximum incense altar bonus from the last incense altar the player has encountered.
(There is a case in which the dagger glows even if the player is not at maximum incense bonus:
 The player must have been at maximum incense bonus and then gone into the vincinity of a stronger incense altar.
 The maximum incense bonus data field only updates once the maximum bonus has been reached for efficiency.)

closes #1034

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/item/ItemSacrificialDagger.java
+++ b/src/main/java/WayofTime/bloodmagic/item/ItemSacrificialDagger.java
@@ -8,10 +8,7 @@ import WayofTime.bloodmagic.item.types.ISubItem;
 import WayofTime.bloodmagic.tile.TileAltar;
 import WayofTime.bloodmagic.util.Constants;
 import WayofTime.bloodmagic.util.DamageSourceBloodMagic;
-import WayofTime.bloodmagic.util.helper.NBTHelper;
-import WayofTime.bloodmagic.util.helper.PlayerHelper;
-import WayofTime.bloodmagic.util.helper.PlayerSacrificeHelper;
-import WayofTime.bloodmagic.util.helper.TextHelper;
+import WayofTime.bloodmagic.util.helper.*;
 import net.minecraft.client.renderer.ItemMeshDefinition;
 import net.minecraft.client.renderer.block.model.ModelResourceLocation;
 import net.minecraft.client.util.ITooltipFlag;
@@ -37,6 +34,8 @@ import java.util.function.Consumer;
 
 public class ItemSacrificialDagger extends ItemEnum<ItemSacrificialDagger.DaggerType> implements IMeshProvider {
 
+    private boolean hasMaxIncense;
+
     public ItemSacrificialDagger() {
         super(DaggerType.class, "sacrificial_dagger");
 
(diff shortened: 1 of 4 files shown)


src/main/java/WayofTime/bloodmagic/item/ItemSacrificialDagger.java
```diff
@@ -37,6 +34,8 @@
 
 public class ItemSacrificialDagger extends ItemEnum<ItemSacrificialDagger.DaggerType> implements IMeshProvider {
 
+    private boolean hasMaxIncense;
```
TehNut: Nope not how this works.

TehNut (changes requested):

src/main/java/WayofTime/bloodmagic/item/ItemSacrificialDagger.java
Arcaratus: :face_palm:

keraldi: Well, how DOES it work then?
I take it you mean: "This is not how/where we declare booleans."
or "Don't use a boolean for this here, use something else to compare/store 2 NBT tags that you check against each other."

Because it does what I intend it to do and as efficient as I can imagine it to do it.

What SHOULD I use?
Where should I put it?

TehNut: Items have a single registered instance.

keraldi: So... static?
Didn't know, know it now for the future.

UpcraftLP: no, not static. Items are flyweights, so making the field static does not change anything, as instance variables are effectively static.

keraldi: It's great that you all tell me what I shouldn't do but what should I do instead of that what I shouldn't do?

Because simply saying what should not be done doesn't change the fact that I didn't know better in that case and probably didn't find it out on my own while I was doing what I ended up doing.

keraldi:
Finished. Lesson learned :P

src/main/java/WayofTime/bloodmagic/util/Constants.java
```diff
@@ -129,6 +131,7 @@
         public static final String TANK = "tank";
 
         public static final String BREATH = "breath";
+        public static final String MAXINCENSE = "BM:";
```
TehNut: ?

TehNut (changes requested):

TehNut (approved):

PR #1394, opened 2018-08-07
Self-Sacrifice regen nerf ( Plain sacrificial dagger only)

### LINKED ISSUE #579: [Idea] Regen Problem
As I've heard direwolf20 say countless times "WayOfTime doesn't like people using regen at the altar".
So I was thinking; why not check for regen effect, then take more LP with a low chance to kill the player(all LP).

keraldi:
Self-Sacrifice nerf for unimbued (no tranquility) sacrifices while affected by Regeneration.

(Imbued sacrifices already have a cooldown with Soul Fray which makes nerfing them based on Regeneration pointless.)
Didn't exactly calculate through but it should still give a slight net positive if sacrificing constantly with Regeneration.

based on #579

closes #579


TehNut:
No

Issue #1382, opened 2018-07-31
Blood Tank does not work with Water/Lava/Void Sigil


keraldi:
## Issue Description:
Trying to add Water/Lava/Void to a Blood Tank with the corresponding Sigil does not work.

## What happens:
Water/Lava/Void Sigil do not add the corresponding "stuff" to the Blood Tank.

## What you expected to happen:
"Stuff" should be added to the Blood Tank, just as it does for (most) other liquid containers.

## Steps to reproduce:

1. Get "stuff" Sigils
2. Bind them
3. Use them on the Blood Tank
(4. (For reference) Use them on a forge liquid container from another mod)
...

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: 1.12.2-2.3.1-99 (compiled from source)
- Minecraft: 1.12.2
- Forge: 2611 (ran with IDE default settings)

TehNut:
Should be fixed by #1406

Issue #1301, opened 2018-05-11
Lava sigil dupes lava when right clicked on some tanks


justinrusso:
#### Issue Description:
When right-clicking some tanks, the lava sigil fills the tank as well as places a lava fluid block outside of the tank where right-clicked. Reference https://github.com/DarkPacks/SevTech-Ages/issues/2059 and https://github.com/DarkPacks/SevTech-Ages/issues/1781

#### What happens:
I've tested on the following and categorized into what worked and what did not:

Does not dupe:

    IE Tank
    Ender Tank

Dupes:

    BC Tank
    Tinkers Seared Tank
    Cyclic Fluid Storage Tank

#### What you expected to happen:
No duping

#### Steps to reproduce:

1. Place one of the tanks down listed above that dupes
2. Right-click the tank with a lava sigil

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: 1.12.2-2.2.12-97
- Minecraft: 1.12.2
- Forge: 14.23.3.2681

xenxero1:
When this happens, does this consume the normal amount of LP, or double?

justinrusso:
Seems to have used 2000 LP when I tested just now, but the OP from the SevTech issue states it used 1000. So not a dupe on my end? but still annoying.

DenisMasterHerobrine:
Did some tests, with tanks and Lava Sigil and tried to reproduce it and got the some thing. If you only Righ-Click to a tank, it will consume 2000 LP and you get 2 buckets, but if you Shift + Right-Click to a tank with a sigil, it'll consume 1000 LP, and you also get 2 buckets. 

Tested versions:
BloodMagic 2.2.7-90 & 2.2.12-97
Minecraft: 1.12.2
Forge: 14.23.3.2681

PR #1406, opened 2018-08-17
Changed the lava, water, and void sigils to work with fluid tanks better

ChrisThirtle:
This addresses the issues numbered #1382 and #1301 as well as a minor issue of total incompatibility with certain kinds of tanks by making the water, lava, and void sigils largely use Forge's internal FluidUtils and more strongly implementing the IFluidHandler interface for everything they do.
Of note for anyone testing this out, there is an issue with Blood Tanks in that they do not visually update when used with a sigil. But even if they don't visually update, Blood Tanks _do_ receive the respective fluids from the sigils. This can be seen by using a sigil on a blood tank and breaking it with a pickaxe.

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/item/sigil/ItemSigilFluidBase.java
+++ b/src/main/java/WayofTime/bloodmagic/item/sigil/ItemSigilFluidBase.java
@@ -1,0 +1,120 @@
+package WayofTime.bloodmagic.item.sigil;
+
+import net.minecraft.block.Block;
+import net.minecraft.block.BlockLiquid;
+import net.minecraft.block.material.Material;
+import net.minecraft.block.state.IBlockState;
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.tileentity.TileEntity;
+import net.minecraft.util.EnumFacing;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+import net.minecraftforge.fluids.IFluidBlock;
+import net.minecraftforge.fluids.FluidStack;
+import net.minecraftforge.fluids.Fluid;
+import net.minecraftforge.fluids.capability.CapabilityFluidHandler;
+import net.minecraftforge.fluids.capability.wrappers.BlockLiquidWrapper;
+import net.minecraftforge.fluids.capability.wrappers.BlockWrapper;
+import net.minecraftforge.fluids.capability.wrappers.FluidBlockWrapper;
+import net.minecraftforge.fluids.capability.IFluidHandler;
+
+public abstract class ItemSigilFluidBase extends ItemSigilBase {
+    //Class for sigils that interact with fluids, either creating or deleting them.
+    //Sigils still have to define their own onRightClick behavior, but the actual fluid-interacting code is largely limited to here.
+    public final FluidStack sigilFluid;
+    
+    public ItemSigilFluidBase(String name, int lpUsed, FluidStack fluid){
+        super(name, lpUsed);
+        sigilFluid = fluid;
+    }
+    public ItemSigilFluidBase(String name){
+        super(name);
+        sigilFluid = null;
+    }
+    
+    //The following are handler functions for fluids, all genericized.
+    //They're all based off of the Forge FluidUtil methods, but directly taking the sigilFluid constant instead of getting an argument.
+    
+    /* Gets a fluid handler for the targeted block and siding.
+     * Works for both tile entity liquid containers and fluid blocks.
+     * This one is literally identical to the FluidUtil method of the same signature.
+     */
+    protected IFluidHandler getFluidHandler(World world, BlockPos blockPos, EnumFacing side) {
+        IBlockState state = world.getBlockState(blockPos);
+        Block block = state.getBlock();
+        
+        if (block.hasTileEntity(state)) {
+            TileEntity tile = world.getTileEntity(blockPos);
+            if(tile != null && tile.hasCapability(CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY, side))
+                return tile.getCapability(CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY, side);
+        }
+        if (block instanceof IFluidBlock)
+            return new FluidBlockWrapper((IFluidBlock) block, world, blockPos);
+        else if (block instanceof BlockLiquid)
+            return new BlockLiquidWrapper((BlockLiquid) block, world, blockPos);
+        return null;
+    }
+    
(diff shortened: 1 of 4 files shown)


src/main/java/WayofTime/bloodmagic/item/sigil/ItemSigilLava.java
```diff
@@ -34,80 +41,61 @@ public ItemSigilLava() {
...
-
-                    if (!player.canPlayerEdit(blockpos1, rayTrace.sideHit, stack)) {
-                        return super.onItemRightClick(world, player, hand);
-                    }
-
-                    if (canPlaceLava(world, blockpos1) && NetworkHelper.getSoulNetwork(getBinding(stack)).syphonAndDamage(player, SoulTicket.item(stack, world, player, getLpUsed())).isSuccess() && tryPlaceLava(world, blockpos1)) {
-                        return super.onItemRightClick(world, player, hand);
+            
+            ActionResult<ItemStack> ret = ForgeEventFactory.onBucketUse(player, world, stack, rayTrace);
+            if (ret != null) return ret;
```
TehNut: Won't this bypass syphoning for placing lava in world?

TehNut (changes requested):
`IFluidHandlerItem` is a capability interface, it's not intended to be implemented on objects like this.

src/main/java/WayofTime/bloodmagic/item/sigil/ItemSigilLava.java
ChrisThirtle: I've never actually had the onBucketUse event thing fire in all of my testing, since it still runs off to the actual FluidUtil#tryPlaceFluid later on and places the fluid with the syphon code. I'll remove that snippet just to be sure though.

If I can't implement IFluidHandlerItem due to the fact that actually giving the sigils the correct hasCapability causes their ability to syphon on draining to actually break (due to the game apparently favoring running the tile entity's on click handlers it seems, at least when you're not shift clicking), then I'll have to dump everything that uses the FluidUtil methods and essentially re-implement the ones each sigil uses in the sigils themselves (with a few modifications like not trying to drain from the sigils).
So if the standard is strict and I definitely should not implement IFluidHandlerItem without also declaring hasCapability/getCapability for FLUID_HANDLER_ITEM_CAPABILITY, I will do so. Just wanna be sure that that's the only issue at hand and I'm not missing something else on that subject.

ChrisThirtle:
I made ItemSigilFluidBase as a class to hold the common logic and direct fluid-interacting code for the 3 fluid sigils, replacing the FluidUtil functions, and in doing so made the sigils not need to implement IFluidHandler themselves. I feel it's better to keep the fluid code in one place because if it's broken or if it breaks, it's easier to maintain it in one location than two.
So all the sigils are reimplemented as "extends ItemSigilFluidBase" (which is just an extension of ItemSigilBase itself), and they simply define a FluidStack in their constructor for whatever kind of fluid they create. Then they can use any of the functions from the new class to do all their fluid activities.

src/main/java/WayofTime/bloodmagic/item/sigil/ItemSigilFluidBase.java
```diff
@@ -0,0 +1,120 @@
...
+     * Works for both tile entity liquid containers and fluid blocks.
+     * This one is literally identical to the FluidUtil method of the same signature.
+     */
+    protected IFluidHandler getFluidHandler(World world, BlockPos blockPos, EnumFacing side) {
+        IBlockState state = world.getBlockState(blockPos);
+        Block block = state.getBlock();
+        
+        if (block.hasTileEntity(state)) {
+            TileEntity tile = world.getTileEntity(blockPos);
+            if(tile != null && tile.hasCapability(CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY, side))
```
TehNut: While it may make sense to normal people to do it like this, you're *apparently* not supposed to. This should be something like

```java
if (tile != null) {
    IFluidHandler fluidHandler = tile.getCapability(CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY, side);
    if (fluidHandler != null)
        return fluidHandler;
}
```

src/main/java/WayofTime/bloodmagic/item/sigil/ItemSigilFluidBase.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+    
+    //The following are handler functions for fluids, all genericized.
+    //They're all based off of the Forge FluidUtil methods, but directly taking the sigilFluid constant instead of getting an argument.
+    
+    /* Gets a fluid handler for the targeted block and siding.
+     * Works for both tile entity liquid containers and fluid blocks.
+     * This one is literally identical to the FluidUtil method of the same signature.
+     */
+    protected IFluidHandler getFluidHandler(World world, BlockPos blockPos, EnumFacing side) {
```
TehNut: `@Nullable`

src/main/java/WayofTime/bloodmagic/item/sigil/ItemSigilLava.java
```diff
@@ -54,48 +49,27 @@ public ItemSigilLava() {
...
-                IFluidHandler destinationSide = FluidUtil.getFluidHandler(world, blockPos, rayTrace.sideHit);
-                if(destinationSide != null && FluidUtil.tryFluidTransfer(destinationSide, this, getFluid(), false) != null && NetworkHelper.getSoulNetwork(getBinding(stack)).syphonAndDamage(player, SoulTicket.item(stack, world, player, getLpUsed())).isSuccess()) {
-                    FluidStack result = FluidUtil.tryFluidTransfer(destinationSide, this, this.getFluid(), true);
-                    if (result != null) return ActionResult.newResult(EnumActionResult.SUCCESS, stack);
+                IFluidHandler destinationSide = getFluidHandler(world, blockPos, rayTrace.sideHit);
+                if(destinationSide != null && tryInsertSigilFluid(destinationSide, false) && NetworkHelper.getSoulNetwork(getBinding(stack)).syphonAndDamage(player, SoulTicket.item(stack, world, player, getLpUsed())).isSuccess()) {
+                    boolean result = tryInsertSigilFluid(destinationSide, true);
+                    if (result) return ActionResult.newResult(EnumActionResult.SUCCESS, stack);
```
TehNut: newlines on all returns like this

TehNut (changes requested):

ChrisThirtle:
Did the above stuff and I'm starting to get everything set up in the IDE to actually make sure my future code is more consistent.

TehNut (approved):

PR #1408, opened 2018-08-20
Update ko_KR.lang

mindy15963:


TehNut:
1.7.10 is no longer supported and we are not accepting PRs for it anymore. If you would like to re-submit this for 1.12, I would be happy to merge it.

PR #1424, opened 2018-08-28
Potion tipped & modded arrows now get the will effects applied alongside them.

keraldi:
- Destructive releases a splash potion effect at the target location
 - potion amplifiers are increased if the potion effect is already applied by the will type (poison, levitation, slowness)


TehNut:
I think the more proper fix would be to wrap whatever arrow is being fired and call it's relevant methods where applicable.

keraldi:
That might make it not possible to increase the amplifier if the same effect is applied multiple times.

Edit: nevermind, itemstack could just be passed through for that.

keraldi:
As far as I can see, the problem with wrapping is that almost all relevant stuff is protected or private. If we want to still have will effects alongside Tipped, Spectral or modded arrows, they either need to be created alongside or their effects be read out just like their corresponding entity classes do.

Unless I'm overseeing something or too stupid to see a better path.


EDIT: Trying to use MethodHandle for this (Specifically modded arrows, maybe Tipped/Spectral arrows for potion effect colors).

keraldi:
~~Reverted any interaction with modded arrows.
Sentient Bow now acts as follows:

- Fires Potion Tipped & Spectral arrows with additional will effects (effects will be stronger if it applies the same effect as the will)
- Destructive fired Potion Tipped & Spectral arrows will create a Splash Potion at the hit location.
- Fires Modded arrows without any modification~~

See first post.

keraldi:
Done.

PR #1450, opened 2018-09-25
Fix NullPointerException if no binding.

ginyai:
Check if the binging is null before get soul network.
Check the result of `syphonAndDamage` before SigilBloodLight and SigilTransposition work.


TehNut:
In what cases are these actually problems? By the time any of this is called, there should be a binding available.

ginyai:
https://gist.github.com/ginyai/52d413be10a456d6e0207dbd5e9f235e
Sometimes this happens, but I cannot find out why.
On the other hand, others can cancel the `ItemBindEvent` to deny the binding.

keraldi:
Did you ever bind a blood orb (created a soul network) when doing that?

ginyai:
@Iorce I did. And I think it is irrelevant to the exception.

Issue #910, opened 2016-09-10
Feature request:  "Unbound" warning in tooltips


MentalMouse:
After the N'th time forgetting to bind something, I'm making it a feature request:  That the tooltip for anything that requires binding and isn't currently bound, should explicitly say "Unbound".  Ideally in colored text.

ETA:  And, no, the absence of a bit of text which is otherwise so common as to be "invisible", is not nearly as informative as an explicit warning.

keraldi:
Unwanted. Invalid. Close. (I actually implemented it and Nut rejected it (The PR is still open BTW, changed to your liking)).

PR #1455, opened 2018-10-02
Removed some redundant stuff for Bindable items, improved inheritance

keraldi:
- Streamlined inheritance. All Sigils now extend ItemBindableBase at some point.

- removed redundant hasTagCompound check

- removed usage of TextHelper in affected files

#910 (Nevermind this, it's a tooltip request)

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/item/ItemBindableBase.java
+++ b/src/main/java/WayofTime/bloodmagic/item/ItemBindableBase.java
@@ -3,7 +3,6 @@ package WayofTime.bloodmagic.item;
 import WayofTime.bloodmagic.BloodMagic;
 import WayofTime.bloodmagic.core.data.Binding;
 import WayofTime.bloodmagic.iface.IBindable;
-import WayofTime.bloodmagic.util.helper.TextHelper;
 import net.minecraft.client.util.ITooltipFlag;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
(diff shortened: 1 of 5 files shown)


TehNut:
I don't particularly agree with the unbound tooltip.

keraldi:
As in:
- Unnecessary (the whole PR, close it)
- Could be realized better
- Remove the tooltip thing (because everyone who forgets binding is at their own fault) and keep everything else?

TehNut:
The last one.

src/main/java/WayofTime/bloodmagic/item/ItemBindableBase.java
```diff
@@ -24,11 +24,11 @@ public ItemBindableBase() {
     @Override
     @SideOnly(Side.CLIENT)
     public void addInformation(ItemStack stack, World world, List<String> tooltip, ITooltipFlag flag) {
-        if (!stack.hasTagCompound())
-            return;
-
-        Binding binding = getBinding(stack);
-        if (binding != null)
-            tooltip.add(TextHelper.localizeEffect("tooltip.bloodmagic.currentOwner", binding.getOwnerName()));
+        if (stack.hasTagCompound()) {
```
TehNut: There is no reason for these formatting changes

TehNut (changes requested):

src/main/java/WayofTime/bloodmagic/item/ItemBindableBase.java
keraldi: I made it to prevent code duplication when I added a tooltip, I'll revert it now since said tooltip is not relevant anymore.

TehNut (approved):

Issue #1457, opened 2018-10-03
Demon Crystal Cluster causing serious render lag


tommyTT:
#### Issue Description:
Having Demon Crystal Cluster grow in a chunk causes serious render lag after a recent update to the mod.

#### What happens:
Whenever you are in the vicinity of a crystal cluster the game experiences those micro stutters in the rendering and everything isn't smooth any more. If you remove the cluster everything immediately returns to normal.

#### What you expected to happen:
The cluster shouldn't affect the rendering that much, especially since it happens constantly and not only when a cluster actually grows.

#### Steps to reproduce:
1. Put down a crystal cluster in a chunk.
2. Look around in the world when near the cluster. Maybe even pull up the render graph (I'm using the Sampler mod for that).
3. Now remove the cluster and repeat step 2.

#### Analysis
I looked a bit further into the code since this only became an issue in a recent version of the mod and found [this ](https://github.com/WayofTime/BloodMagic/blob/1.12/src/main/java/WayofTime/bloodmagic/tile/TileDemonCrystal.java#L39) addition to the tile entity code of the cluster. When I uncomment this update call in my dev environment, everything is smooth again but the original problem of course returns. Forcing a render update every 20 ticks even if the cluster didn't grow causes those render issues and isn't the solution to the problem. Looking further into the code I found that the [growth](https://github.com/WayofTime/BloodMagic/blob/1.12/src/main/java/WayofTime/bloodmagic/tile/TileDemonCrystal.java#L134) method tries to cause that update already but since the block state never actually changes, the same block state gets syncronized to the client again, therefore not updating the rendering. I then looked into how [Astral Sorcery](https://github.com/HellFirePvP/AstralSorcery/blob/master/src/main/java/hellfirepvp/astralsorcery/common/tile/TileCelestialCrystals.java#L124) e.g. does the crystal growth and changed that method to look like this:
```java
public void checkAndGrowCrystal() {
        if (progressToNextCrystal >= 1 && internalCounter % 100 == 0) {
            progressToNextCrystal--;
            crystalCount++;

            IBlockState state = getWorld().getBlockState(pos);
            IBlockState newState = state.getActualState(getWorld(), pos);
            getWorld().setBlockState(pos, newState);
      }
 }
```
Now everything updates like it should when a cluster grows without causing those constant render updates. Since I'm no expert at Modding and especially rendering, I don't know if this is a good solution for this mod. It should however demonstrate the underlying problem that was the intent of that bugfix.
____
[...]

PR #1458, opened 2018-10-03
Bugfix attempt to remove render lag (see #1457)

tommyTT:
See #1457 for details and concerns that should be adressed before merging.

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/tile/TileDemonCrystal.java
+++ b/src/main/java/WayofTime/bloodmagic/tile/TileDemonCrystal.java
@@ -35,8 +35,6 @@ public class TileDemonCrystal extends TileTicking
     {
         if (getWorld().isRemote)
         {
-            if(internalCounter % 20 == 0)
-                getWorld().markBlockRangeForRenderUpdate(pos, pos);
             return;
         }
 
@@ -207,5 +205,11 @@ public class TileDemonCrystal extends TileTicking
     public void setPlacement(EnumFacing placement)
     {
         this.placement = placement;
+    }
+
+    @Override
+    protected void onDataPacketClientReceived() {
+        super.onDataPacketClientReceived();
+        notifyUpdate();
     }
 }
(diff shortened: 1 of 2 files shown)


keraldi:
While naturally growing the crystal updates it correctly and immediately, removing the crystal by right clicking the cluster (possible while carrying a Tartaric Gem with > 1024 of Raw Will) does not update it.

Same goes for adding crystals manually through creative mode.

Please edit https://github.com/tommyTT/BloodMagic/blob/2e9ad1122b0816f22731f7e75b91a6ade0be40af/src/main/java/WayofTime/bloodmagic/block/BlockDemonCrystal.java#L50 accordingly.

While markdirty() and notifyUpdate() might be meaningless for the visual update, eventListeners should probably still know if something happened there, though @TehNut's say will have more weight on that.

TehNut:
`markDirty()` is called to inform the world that a value has changed and should be written to disk during the next save. Since values are changing, it needs to be called.

tommyTT:
Sorry for the delay, I looked more into the underlying problem and found a more general solution that doesn't require to change 10 places where the tile entity can change.

This now applies to all TileEntities because the method I changed was final. I think this is fine, but you might need to verify and adapt the code accordingly.

keraldi:
Works for me.

This is now obsolete and should be removed:

https://github.com/WayofTime/BloodMagic/blob/348a740c14eb3cefe4e377757a4fec93ea6d88c7/src/main/java/WayofTime/bloodmagic/tile/TileDemonCrystal.java#L133

tommyTT:
I'm not 100% sure and therefore didn't change it, since it is called server side. The notifyUpdate I added is called client side.

keraldi:
Try running your change with the "Crack of the Fractured Crystal" Ritual.

Nvm. Inserted the change at the wrong location :P

tommyTT:
Okay, it seems to update correctly for me.

keraldi:
I tested it as well. Does what it intends to do. No noticable side effects.

Tested in: SSP, local SMP with relevant rituals (rituals that grow or destroy/pop crystals, manual adding and removing crystals).

tommyTT:
The previous solution did fix the issue with the crystals but caused significant lag with the blood altar. The new solution now leaves it up to the tile entity to do the required updates. While extensively testing in SSP and SMP I didn't notice any side effects and the crystal growth issue is also fixed.

keraldi:
Please clean up the PR by cherry-picking the changes you made and putting them into a new branch from upstream.

The file changes dialog shows many changes that had absolutely nothing to do with your original PR.

(i.e. close the PR, branch from upstream, paste your changes, open new PR from the new branch)

tommyTT:
I cleaned it up, so now only the relevant changes should be in the branch. Sorry for the mess.

keraldi (approved):

Issue #1420, opened 2018-08-26
Trickshot fires bonus arrows with a Sentient bow even if the player has no will


keraldi:
#### Issue Description:
Living Armor Trickshot upgrade ignores whether or not an arrow has actually been fired.
Sentient Bows cannot fire arrows if the player has no will in its inventory (no Tartaric Gem or empty Tartaric Gem), however Trickshot arrows are still fired.

#### What happens:
Trickshot fires arrows even if the bow doesn't fire arrows.

#### What you expected to happen:
Trickshot should only fire arrows if the bow fires arrows.

#### Steps to reproduce:

1. Get Trickshot on Living armor
2. Get Sentient Bow (no Tartaric Gem)
3. Use the Bow.
...

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: -100 / compiled from source
- Minecraft: 1.12.2
- Forge: 2611

keraldi:
Bug label

keraldi:
Fixed by #1467

Issue #1421, opened 2018-08-26
Sentient bow does not fire arrows without will but still consumes arrows


keraldi:
#### Issue Description:
Sentient bow consumes arrows but doesn't fire them without will, even in creative mode.
This is inconsistent with the Sentient Sword, which deals damage even if there's no will in the player's inventory.

#### What happens:
Sentient Bow does not fire arrows without will but still consumes them.

#### What you expected to happen:
Sentient Bow should fire arrows if it consumes them.
Sentient Bow should fire arrows even without will (to have consistency with the Sentient Sword.

#### Steps to reproduce:

1. Get Sentient Bow
2. Have no will in the inventory (empty tartaric gem or no tartaric gem)
3. Fire.
...

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: -100 / compiled from source
- Minecraft: 1.12.2
- Forge: 2611

Phil-M-H:
It does not actually consume the arrow. If you pick up your stack of arrows and put it back down, you'll find you will have the same number. It just visually decrements the number of arrows that it displays.

keraldi:
PR in progress.

Phil-M-H:
Resolved in commit 5903cc7 big kek

TehNut:
Actually it was 6b25caa06df3dce862a0fd5e6ea30611d94c8a4e

PR #1466, opened 2018-11-08
Sentient Bow without Demon Will (Issue 1421)

Phil-M-H:
Added a soullessShotVelocity constant for arrow velocity when fired without demon souls. Applied it in the recalculatePowers method, so that if there is no Demon Will (when level >=0 in line 121), it would use the soullessShotVelocity as the arrow velocity.
Modified the onPlayerStoppedUsing function so that it would fire an arrow no matter what, but if there are no demon souls it would play the same fire extinguish sound as before.

Overall result:
If you fire the bow without Demon Will, it will fire at a velocity lower than a normal bow and play the fire extinguished sound. As opposed to what is described in issue 1421 (https://github.com/WayofTime/BloodMagic/issues/1421)


Phil-M-H:
Would also fix #1420

keraldi:
redo the PR. this is a total mess.

PR #1490, opened 2018-12-30
Reverence of the Condor port

keraldi:
Added a small continous cost of `Blockvolume \ 10000 * playerCount`, the `playerCount` variable was always there was no cost. 

Using an AreaDescriptor, the area should be able to be modified. I defined the max. area as 100 horizontal and 200 vertical. Hence the cost based on block volume.

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/ritual/types/RitualCondor.java
+++ b/src/main/java/WayofTime/bloodmagic/ritual/types/RitualCondor.java
@@ -25,7 +25,6 @@ public class RitualCondor extends Ritual {
     @Override
     public void performRitual(IMasterRitualStone masterRitualStone) {
         AxisAlignedBB aabb = getBlockRange(FLIGHT_RANGE).getAABB(masterRitualStone.getBlockPos()).expand(-10, 0, -10);
-        System.out.println(aabb);
         World world = masterRitualStone.getWorldObj();
 
         int currentEssence = masterRitualStone.getOwnerNetwork().getCurrentEssence();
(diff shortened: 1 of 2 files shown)


src/main/java/WayofTime/bloodmagic/ritual/types/RitualCondor.java
```diff
@@ -0,0 +1,90 @@
...
+    public RitualCondor() {
+        super("ritualCondor", 0, 1000000, "ritual." + BloodMagic.MODID + ".condorRitual");
+        addBlockRange(FLIGHT_RANGE, new AreaDescriptor.Rectangle(new BlockPos(0, 0, 0), new BlockPos(10, 30, 10)));
+        setMaximumVolumeAndDistanceOfRange(FLIGHT_RANGE, 0, 100, 200);
+    }
+
+    @Override
+    public void performRitual(IMasterRitualStone masterRitualStone) {
+        AxisAlignedBB aabb = getBlockRange(FLIGHT_RANGE).getAABB(masterRitualStone.getBlockPos()).expand(-10, 0, -10);
+        System.out.println(aabb);
```
TehNut: pls

TehNut (changes requested):

src/main/java/WayofTime/bloodmagic/ritual/types/RitualCondor.java
keraldi: OUPS.

TehNut (approved):

PR #1492, opened 2018-12-31
Ported Ritual of the Feathered Earth to BM2.

keraldi:


TehNut:
I feel like a more user-friendly implementation of this would be to handle the actual effect in a damage event and cancelling it if it's fall damage while they're in range.

This current implementation is just constantly setting fall damage to 0 every second, even for entities which aren't moving, which just seems wrong.

keraldi:
Done.

gradle.properties
```diff
@@ -2,7 +2,7 @@ mod_name=BloodMagic
 package_group=com.wayoftime.bloodmagic
 mod_version=2.3.3
 mc_version=1.12.2
-forge_version=14.23.2.2611
+forge_version=14.23.5.2768
```
TehNut: Is this necessary?

TehNut (changes requested):

gradle.properties
keraldi: My bad.

TehNut (approved):

PR #1501, opened 2019-01-03
 Implemented Ritual of Grounding, a Ritual to change gravity behavior

keraldi:
[x] <- x are new potion effects
 - (NoMod) moves entities towards the ground, prevents jumping [Grounded]
 - (Raw) affects players
 - (Corrosive) disables gravity [Suspension]
 - (Destructive) increases fall damage [Heavy Heart]
 - (Steadfast) affects bosses
 - (Vengeful) stronger effects, (+Corrosive) applies levitation (+Destructive) stronger effect

 [Grounded] prevents jumping and moves entities towards the ground, higher amplifiers cause a faster descend, interesting interaction with Sigil of Air

 [Suspension] disables gravity (keeps movement)

 [Heavy Heart] increases fall height and fall damage multiplier by 1 per level.

 Fixed a possible division by 0 in RitualConder.
 Saved event entity variable in PotionEventHandlers.
 Made rune configuration more readable in RitualHarvest.

Signed-off-by: tobias <angryaeon@icloud.com>


TehNut:
We need to make sure these lists are cleared when the server shuts down. It doesn't matter on dedicated servers, but for clients, this will cause player objects to stick around.

keraldi:
How? Or rather: where?

TehNut:
There's lifecycle events for servers.

keraldi:
What about force shutdowns/crashes/people just shutting power off or force quitting per window manager?

TehNut:
When it closes completely, it's a non-issue since the JVM throws out the objects anyway. What this would do is stop memory leaks by keeping old player objects around while hopping between different saves/servers.

keraldi:
So basically the problem that the RitualReader (Ritual Tinkerer) had (has?)?

TehNut:
Correct, but this causes a leak (old entity objects that also keep old world objects) while that just caused behavioral issues.

keraldi:
`WorldEvent.Unload`

Probably going to use that with world-specific lists (ex. `GenericHandler.bounceMap`) that will then be cleared.

Another possible solution is to clear players from the lists post-tick, however that might be more expensive (need to check for effect again, put on list again etc. every (couple) ticks)

keraldi:
Done.

Would it be possible to make an annotation for it aka. `@ClearOnUnload`? 
Would that be better/more readable?

TehNut:
Have you tested having a client jump between SSP/SMP instances and making sure the map is cleared?

keraldi:
No. SMP -> SSP would not clear the list but SSP -> SMP would.
Values should be saved on the server side (need to check).

TehNut:
Well what about SSP -> SSP

keraldi:
SSP -> SSP triggers a world unload event and therefore clears the list

Unless you crash, in which case objects are dropped anyways.

PR #1517, opened 2019-01-17
Capable fluid sigils

keraldi:
Fluid sigils now act like tanks filled with their respective fluids when put into a machine that supports fluid handlers.


TehNut:
There's a lot of whitespacing that shouldn't be here.

Issue #1475, opened 2018-11-24
Modify cost of Meteor ritual to derive from Configuration


noobanidus:
As discussed with TehNut on discord, potentially modifying the Meteor Ritual to have an activation cost of 0, and determine the actual cost when the player attempts to provide a reagent (as determined by the configuration). 

Filing as requested!

PR #1518, opened 2019-01-18
Mark of the Falling Tower meteor costs are now configurable

keraldi:
via an additonal field "cost" in the .json config files

 if BM is updated from an old version, this would require meteor costs to be added manually for custom meteors; meteors would be for free (LP-wise) otherwise.

 These are the new default costs:
 IRON: 1,000,000 (same as before)
 GOLD: 2,500,000
 DIAMOND: 5,000,000 (requires Archmage Blood Orb)

Default Meteor versions have been increased by 1 in order to update their configuration.

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/item/ItemBoundTool.java
+++ b/src/main/java/WayofTime/bloodmagic/item/ItemBoundTool.java
@@ -244,7 +244,7 @@ public class ItemBoundTool extends ItemTool implements IBindable, IActivatable {
                 if (silkTouch && blockState.getBlock().canSilkHarvest(world, blockPos, world.getBlockState(blockPos), player))
                     ItemHandlerHelper.giveItemToPlayer(player, new ItemStack(Item.getItemFromBlock(blockState.getBlock())));
                 else {
-                    NonNullList<ItemStack> itemDrops = new NonNullList<>();
+                    NonNullList<ItemStack> itemDrops = NonNullList.create();
                     blockState.getBlock().getDrops(itemDrops, world, blockPos, world.getBlockState(blockPos), fortuneLvl);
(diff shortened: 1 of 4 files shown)


keraldi:
#1475

src/main/java/WayofTime/bloodmagic/item/ItemBoundTool.java
```diff
@@ -244,7 +244,7 @@ protected void sharedHarvest(ItemStack stack, World world, EntityPlayer player,
                 if (silkTouch && blockState.getBlock().canSilkHarvest(world, blockPos, world.getBlockState(blockPos), player))
                     ItemHandlerHelper.giveItemToPlayer(player, new ItemStack(Item.getItemFromBlock(blockState.getBlock())));
                 else {
-                    NonNullList<ItemStack> itemDrops = new NonNullList<>();
+                    NonNullList<ItemStack> itemDrops = NonNullList.create();
```
TehNut: Irrelevant change I'll do this myself.

noobanidus:
ilu

src/main/java/WayofTime/bloodmagic/meteor/Meteor.java
```diff
@@ -29,6 +30,7 @@ public Meteor(ItemStack catalystStack, List<MeteorComponent> components, float e
         for (MeteorComponent component : components)
             weight += component.getWeight();
         this.maxWeight = weight;
+        this.cost = 1000000;
```
TehNut: GSON initializes classes with reflection and not the constructor. The default definition should be set at the field declaration.

TehNut (changes requested):

TehNut (approved):

PR #1522, opened 2019-01-21
Fix for version update Ritual NPE

keraldi:
…(not sure if I would have gotten the NPE to begin with)


keraldi:
Eliminates the Ritual NPE issue.

Issue #1526, opened 2019-01-22
Seer's Sigil and Incense Altar


utoxin:
#### Issue Description:
I'm aware that this has been reported before, but that was over 6 months ago, and it seems to be broken still / again.

The Seer's Sigil doesn't provide the on-screen info about the Incense Altar that the Divination Sigil does.

#### What happens:

Divination sigil displays info for both Incense Altar and Blood Altar.
Seer's sigil only displays info for the Blood altar.

#### What you expected to happen:

Both sigils display info for both altars.

#### Steps to reproduce:

1.  Use a Seer's sigil to look at an Incense altar.
____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: 1.12.2-2.3.3-101
- Minecraft: 1.12.2
- Forge: 14.23.5.2781

keraldi:
Can confirm: Seer Sigil does not provide incense altar information

PR #1527, opened 2019-01-22
Seer Sigil can now be used in place of divination sigil in additon to its own functions for HUD elements

keraldi:
(not sure if there might be edge cases that it tries to display both the divination sigil and seer sigil information at the same time (as could be possible for the blood altar), however it doesn't seem to be the case as removing elements from blood_altar_adv removes them from the seer sigil view)

#1526


MentalMouse:
I thought the Seer's Sigil was a straight upgrade of Divination, giving all information provided by the latter?

keraldi:
It's supposed to but there was missing an entry for the Incense Altar HUD element.

Instead of adding a new HUD element for the Seer Sigil that would be identical to the one for the Divination Sigil for the Incense altar, I enabled the Seer Sigil to pass all Divination Sigil checks as well.

MentalMouse:
So basically when it looks at an incense altar, it acts like the Divination sigil?

keraldi:
... Yes.

Issue #1487, opened 2018-12-27
Trickshot arrows to reset hurtresistancetime for bonus damage


keraldi:
Living Armor Trickshot arrows currently do not apply their damage on the desired main target because most of them hit during the hurtresistancetimer.

Important discord excerpt from #general (27.12.2018 15:05 UTC):
```
[4:04 PM] TehNut: Tag trickshot arrows, check for that tag on hit, remove invulnerability
[4:04 PM] AEon: is it a fix though?
[4:04 PM] TehNut: Yes since that's the intended functionality
[4:06 PM] TehNut: Cam always slightly lower the damage :miyashrug:
```

Difficulty: Easy

Phil-M-H:
Resolved in commit 2c49be1

PR #1535, opened 2019-02-07
Trickshot Arrows to reset hurtresistancetime. (#1487)

Phil-M-H:
(#1487)
Added a subscribeevent method for the onProjectileImpact event that
removes the tag and sets the hurtresistanttime to zero.

Allows all trickshot arrows to hit the target, rather than all but the 1st bouncing off.

[lines 164 and 167 were changed so that their formatting would align with how the rest of the file was already formatted]


src/main/java/WayofTime/bloodmagic/util/handler/event/LivingArmourHandler.java
```diff
@@ -341,7 +344,7 @@ public static void onArrowFire(ArrowLooseEvent event)
                         ItemArrow itemarrow = (ItemArrow) ((stack.getItem() instanceof ItemArrow ? arrowStack.getItem() : Items.ARROW));
                         EntityArrow entityarrow = itemarrow.createArrow(world, arrowStack, player);
                         entityarrow.shoot(player, player.rotationPitch, player.rotationYaw, 0.0F, velocity * 3.0F, 1.0F);
-
+                        entityarrow.addTag("Arrow Shot");
```
TehNut: Unless vanilla uses tags named similarly, identifiers like this should be `snake_case`

TehNut (changes requested):

Issue #1415, opened 2018-08-26
Living Armor Soft Fall Upgrade has no effect


gekkobear:
#### Issue Description:
Living Armor Soft Fall Upgrade has no effect.

#### What happens:
You take normal fall damage with Soft Fall Living Armor worn.

#### What you expected to happen:
Reduced fall damage.

#### Steps to reproduce:

1. Get Living Armor with Soft Fall 5
2. Jump off a ledge
3. Take normal fall damage, which should be 100% reduced/prevented.

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: 1.12.2-2.3.1-99.jar
- Minecraft: 1.12.2
- Forge: 14.23.4.2715
Noticed playing Modpack All the Mods 3, upgraded to latest Blood Magic version (listed above); issue still exists.

Socrathustra:
Seconding this bug. I'm playing BloodMagic-1.12.2-2.3.3-101, and it is still not resolved. This is NOT the same as the flickering bug and does not resolve upon restarting the client.

keraldi:
Investigate.

Phil-M-H:
Resolved in commit 28b5caa

PR #1536, opened 2019-02-07
(#1415) Soft Fall / fallProtect Implementation

Phil-M-H:
(#1415)
Fixed softFall / fallProtect upgrade for living armour. Previously it relied on regular armor properties, which fall damage bypasses by being marked is unblockable. 

Created a method subscribed to onPlayerFall which
applies the damage multiplier and the fallProtect stat-tracker. Removed
the armorProperties method of LivingAmourUpgradeFallProtect.java and
replaced it with getDamageMultiplier.


TehNut:
Looks good, though all equations should have spacing (`1 + 1` vs `1+1`) and float declarations should have a capital `F`.

TehNut:
Looks like merging the other PR caused a conflict. Could you resolve that?

Phil-M-H:
Will do

PR #1540, opened 2019-02-13
Trickshot - Sentient Arrows Implementation

Phil-M-H:
Before creating additional arrows, trickshot now checks whether the arrows were shot from a sentient bow. If they were, instead of creating them the typical way it will call a new method, getDuplicateArrow in ItemSentientBow.

getDuplicateArrow is a copy of the method getEntitySentientArrow, without the side effects such as burning up will and arrows. In addition, some of the if statements were removed as their conditions were already checked in LivingArmourHandler before this method would be called.

Total Effect: Trickshot arrows now apply will effects and potion effects. The total will-reimbursement amount is double [If they would get X will, then with any level of trickshot they get 2X will, if every arrow hits]


TehNut:
Please remove the extra linebreaks

Issue #1512, opened 2019-01-13
Destroying Gate of the Fold does not delete its key, thus not being able to create a new Gate with the same key


keraldi:
#### Issue Description:
The Gate of the Fold can only exist up to two times with the same key in any rotation. However, destroying the Ritual (with a bound pickaxe AoE, destroying any block in creative mode or survival mode) does not free up the key which means that a Gate with a now unused key cannot be activated again.

#### What happens:
Rituals with previously used and now unused keys cannot be activated.

#### What you expected to happen:
Be able to activate Rituals that share an unused key.

#### Steps to reproduce:

1. Build 2 Gate of the Fold rituals with the same configuration and activate them
2. Destroy / deactivate any number of existing Gates
3. Build a 3rd Gate with the same configuration as 2 previously existing Gates of the Fold
4. Attempt to activate Gate of the Fold (from step 3)
5. Ritual does not activate 
...

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: -101 (source)
- Minecraft: 1.12.2
- Forge: (source)

wrincewind:
Closing as 1.12 is out of support. Looks like this may have been fixed in the interim (see above commits)

PR #1553, opened 2019-03-02
Gate of the fold Fixes and minor clean up

Phil-M-H:
1. Removing a PortalLocation did not function properly, resulting portals teleporting to already destroyed portals. Fixed with line 175 of RitualPortal

2. The 1st portal added into a world would not function properly, So when you created your first 2 portals and tried to initialize one, then the other, they wouldn't work, and you would instead have to use the activation crystal on the 1st one a second time. 

Fixed making the LocationsHandler properly static. I was unsure of what the getLocationsHandler() method was supposed to do - I'm assuming to verify that the various variables had already been loaded correctly from the file, so I tried to keep that functionality with the verifyIsInitialized() and the initialized boolean. 

3. Having multiple gates was extremely inconsistent, with some of the previous gates still having their portal tiles but not being functional. I changed the functionality so that all gates teleport to the most recently activated gate with the same key. So if you have 5 gates, they will all teleport to the most recently activated gate. The most recently activated gate will teleport to the 2nd most recently activated gate.


keraldi:
#1512

src/main/java/WayofTime/bloodmagic/block/BlockDimensionalPortal.java
```diff
@@ -76,27 +76,24 @@ public void onEntityCollision(World world, BlockPos pos, IBlockState blockState,
...
-                        }
+            LocationsHandler.verifyIsInitialized();
+            ArrayList<PortalLocation> linkedLocations = LocationsHandler.getLinkedLocations(tile.portalID);
+
+            if (linkedLocations != null && linkedLocations.size() > 1) {
+                if (world.getTileEntity(tile.getMasterStonePos()) != null && world.getTileEntity(tile.getMasterStonePos()) instanceof IMasterRitualStone) {
+                    IMasterRitualStone masterRitualStone = (IMasterRitualStone) world.getTileEntity(tile.getMasterStonePos());
+                    PortalLocation portal;
+                    int index = linkedLocations.size() - 1; //index of most recent PortalLocation
+                    if (linkedLocations.get(index).equals(new PortalLocation(masterRitualStone.getBlockPos().up(), world.provider.getDimension()))) {
```
TehNut: https://github.com/WayofTime/BloodMagic/blob/4c76d42ef4a0cd37e70df38acde31e8d0542dfff/src/main/java/WayofTime/bloodmagic/block/BlockDimensionalPortal.java#L87-L91

-->

```java
if (linkedLocations.get(index).equals(new PortalLocation(masterRitualStone.getBlockPos().up(), world.provider.getDimension())))
    index--; // if most recent PortalLocation = this, get the 2nd most recent
portal = linkedLocations.get(index);
```

src/main/java/WayofTime/bloodmagic/block/BlockDimensionalPortal.java
```diff
@@ -76,27 +76,24 @@ public void onEntityCollision(World world, BlockPos pos, IBlockState blockState,
...
-                                TeleportQueue.getInstance().addITeleport(new Teleports.TeleportSameDim(portal.getX(), portal.getY(), portal.getZ(), entity, masterRitualStone.getOwner(), false));
-                            } else {
-                                TeleportQueue.getInstance().addITeleport(new Teleports.TeleportToDim(portal.getX(), portal.getY(), portal.getZ(), entity, masterRitualStone.getOwner(), world, portal.getDimension(), false));
-                            }
-                        }
+            LocationsHandler.verifyIsInitialized();
+            ArrayList<PortalLocation> linkedLocations = LocationsHandler.getLinkedLocations(tile.portalID);
+
+            if (linkedLocations != null && linkedLocations.size() > 1) {
+                if (world.getTileEntity(tile.getMasterStonePos()) != null && world.getTileEntity(tile.getMasterStonePos()) instanceof IMasterRitualStone) {
```
TehNut: `instanceof` implies nonnull. The tile obtained here should also be stored to avoid unnecessary lookups

src/main/java/WayofTime/bloodmagic/block/BlockDimensionalPortal.java
```diff
@@ -76,27 +76,24 @@ public void onEntityCollision(World world, BlockPos pos, IBlockState blockState,
...
-                            }
-                        } else if (linkedLocations.get(1).equals(new PortalLocation(masterRitualStone.getBlockPos().up(), world.provider.getDimension()))) {
-                            PortalLocation portal = linkedLocations.get(0);
-                            if (portal.getDimension() == world.provider.getDimension()) {
-                                TeleportQueue.getInstance().addITeleport(new Teleports.TeleportSameDim(portal.getX(), portal.getY(), portal.getZ(), entity, masterRitualStone.getOwner(), false));
-                            } else {
-                                TeleportQueue.getInstance().addITeleport(new Teleports.TeleportToDim(portal.getX(), portal.getY(), portal.getZ(), entity, masterRitualStone.getOwner(), world, portal.getDimension(), false));
-                            }
-                        }
+            LocationsHandler.verifyIsInitialized();
```
TehNut: Probably better to call this once when the server loads rather than in 3 different places.

TehNut (changes requested):

src/main/java/WayofTime/bloodmagic/block/BlockDimensionalPortal.java
Phil-M-H: How should I go about that, an event handler for serverStarting, or put it in GenericHandler's  onWorldLoad?

src/main/java/WayofTime/bloodmagic/ritual/types/RitualPortal.java
```diff
@@ -172,7 +170,7 @@ public void stopRitual(IMasterRitualStone masterRitualStone, BreakType breakType
             }
         }
 
-        portalRitualTag.removeTag(PORTAL_ID_TAG);
+        portalRitualTag.removeTag(portalRitualTag.getString(PORTAL_ID_TAG));
```
TehNut, line 173: I keep looking at this and I can't help but feel it's wrong. Does `removeTag` not take the key?

TehNut (approved):

src/main/java/WayofTime/bloodmagic/ritual/types/RitualPortal.java
Phil-M-H: It does not. That was what probably was the most annoying to track down, because it just wouldn't remove previous portal locations.

A-UNDERSCORE-D:
you have an extra semicolon at https://github.com/WayofTime/BloodMagic/pull/1553/files#diff-81ae3006fe3f9b05ca3f89855c43a9ceR471

keraldi:
Still broken.
With multiple portals set up, they all direct to the one created last.

Phil-M-H:
> Still broken.
> With multiple portals set up, they all direct to the one created last.

I designed it so that portals with identical keys would all teleport to the last portal with that key created -- and the last portal with that key would direct to the second to last.

It is less buggy now, but it still needs additions like factoring in all sides that touch the portal, and keeping track of the player's direction with respect to the portal. Probably another two weeks or more until I will have the time, unfortunately.

keraldi (approved):

Issue #1444, opened 2018-09-15
Sacrificial Dagger bypasses Death Save mechanisms


qrss:
#### Issue Description:
Grim Reaper's Sprint on Living Armor does not save you from killing yourself with a Sacrificial Dagger.
This behavior of the Sacrificial Dagger might be causing more serious issues with Death Save mechanisms from other mods: TeamTwilight/twilightforest#625

#### What happens:
Grim Reaper's Sprint triggers, but you still die.

#### What you expected to happen:
Grim Reaper's Sprint saves you.

#### Steps to reproduce:
1. Wear Living Armor with the Grim Reaper's Sprint upgrade.
2. Kill yourself with a Sacrificial Dagger.

____
#### Affected Versions:

- BloodMagic: 2.3.3-101
- Minecraft: 1.12.2
- Forge: 14.23.4.2760

qrss:
On Twilight Forest's issue tracker it was suggested that replacing the code
```
player.onDeath(DamageSourceBloodMagic.INSTANCE);
player.setHealth(0);
```
with
```
player.setHealth(0);
player.onDeath(DamageSourceBloodMagic.INSTANCE);
```
in the function onItemRightClick in the source file ItemSacrificialDagger.java (lines 101 and 102) might solve this problem.

Yamza:
Similar issue with Thaumcraft's Charm of Undying
https://github.com/AllTheMods/ATM3-Remix/issues/79

Iorce:
Update?

keraldi:
Update. Enhancement/Suggestion. Bug label?

Issue #1351, opened 2018-06-24
Blood Magic Dagger of Sacrifice causing abnormal death process


MajorGeneralRelativity:
#### Issue Description:
This does occur in a modpack, but it appears to potentially be a mod interaction problem at the very least, so it's fitting. When you kill yourself using a Dagger of Sacrifice, an abnormal death process occurs, and you do not have your inventory saved by OpenBlocks.

#### What happens:
Upon right clicking with the Dagger of Sacrifice one too many times, you die. However, a grave does not generate, and OpenBlocks does not save a copy of your inventory to restore, which leads to your items not being saved at all, nor dumped out in the world like in the absence of any other mods (and keepInventory)

#### What you expected to happen:
OpenBlocks to spit out a "if you want to restore your inventory, use command blah-blah-death-0 to restore your inventory" and generate a grave.

#### Steps to reproduce:

1. Install Modpack at: http://bit.ly/2K7ueRO
2. Kill yourself by holding right click with Dagger of Sacrifice
3. Profit?
...

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: BloodMagic-1.12.2-2.2.12-97.jar (latest 😉 )
- Minecraft: 1.12.2
- Forge: 14.23.4.2705

Please let me know if I need to provide any further information

keraldi:
Why are you using a link shortener (with possible advertisements even) on a website that most people will access from a non-mobile OS?
It doesn't add to convenience and harbors the potential for malware on the other side.

Link destination: https://drive.google.com/file/d/1yN0sqyhsOxNxbNS8HcJqcByhcsRHmC-q

In case you copied the link straight out from a page featuring the modpack, you can use one of the following tools in the future to "unshorten" a link:

https://unshorten.link/ (full HTTPS)
http://redirectdetective.com/
https://checkshorturl.com/
http://getlinkinfo.com/

In either case you probably want to remove the clutter after the actual link that refers to your browser, OS or other stuff.

MajorGeneralRelativity:
I'm the modpack author, so I know that the link is advertisement free and safe. The link I provided is the link I give to everyone on my Discord to download the modpack. I was unaware it would be an issue, as it wasn't mentioned in the contribution guidelines, or the template.

keraldi:
Needs to be investigated. Update. Bug label.

PR #1554, opened 2019-03-02
Possible fix to (many of the) abnormal death events concerning the sacrificial Dagger

keraldi:
=> #1444

Works with Grim Reaper's Sprint, not tested for anything else.
Basically a glorified /kill.


keraldi:
#1351

Issue #1533, opened 2019-02-02
[Suggestion] Infinite/Creative items


keraldi:
For creative worlds, adventure worlds, testing et al

creative/infinte versions of:

- Tartaric Gems (specifically useful for rituals in spawning regions etc)
- Creative Blood Orb (port from 1.7.10, makes commands for filling/emptying etc obsolete for testing)
- Living Armor chestplate with high amount of points (1000 / 10000 ?)
- Other stuff?

keraldi:
- creative activiation crystal makes rituals free (both continous cost and activation) regardless of gamemode

keraldi:
- creative ritual tinkerer (able to ignore boundaries)?

PR #1556, opened 2019-03-03
Added creative handling and/or creative variants of items.

keraldi:
Creative handling:
 - Ritual Diviner: instantly places the ritual, ignores existing blocks (they will be overwritten)
 - Draft of Angelus:
    - applies a limit of 10,000 upgrade points to the living armor if the living armor already has 200 or more (less than 10,000) upgrade points
    - prevented adding of negative effects to the player if the player is in creative mode
 
Creative items:
 - Tartaric Gems: Integer.MAX_VALUE capacity, draining function is disabled.
 - Orb of Testing: Shift+Rightclick -> fill, Rightclick -> cap
                   Shift+Leftclick -> empty, Leftclick -> get
 - Creative Activation Crystal:
   - Rituals won't drain LP anymore if activated by a Creative Activation Crystal (sets soul ticket amount to 0)
     - This change makes "Cry of the Eternal Soul" a steady source of LP income if activated by a Creative Activation Crystal
   - Rituals won't cause Nausea if activated by a Creative Activation Crystal
     - If they don't have any other terminating condition, they will function as if supplied with everything they need
   - Rituals won't drain demonic Will (ethereal/aura) anymore if activated by a Creative Activation Crystal. Since most (all) rituals that use Will have checks in place if there is enough Will to continue, they still require Will to apply Will effects though.

Misc:
 - replaced all occurrences of ".getDisplayName().getFormattedText()" with ".getDisplayName()" whenever an ITextComponent would be possible.
 - fixed translation keys having typos

#1533


src/main/java/WayofTime/bloodmagic/item/soul/ItemSoulGem.java
```diff
@@ -32,7 +32,7 @@
 import java.util.function.Consumer;
 
 public class ItemSoulGem extends Item implements IDemonWillGem, IMeshProvider, IMultiWillTool {
-    public static String[] names = {"petty", "lesser", "common", "greater", "grand"};
+    public static String[] names = {"creative", "petty", "lesser", "common", "greater", "grand"};
```
TehNut: Ah, yes, let's just turn every existing petty gem into a creative one.

src/main/java/WayofTime/bloodmagic/item/alchemy/ItemLivingArmourPointsUpgrade.java
```diff
@@ -63,6 +63,14 @@ public ItemStack onItemUseFinish(ItemStack stack, World worldIn, EntityLivingBas
                 ItemStack chestStack = Iterables.toArray(player.getArmorInventoryList(), ItemStack.class)[2];
                 LivingArmour armour = ItemLivingArmour.getLivingArmour(chestStack);
                 if (armour != null) {
+                    if (player.capabilities.isCreativeMode) {
```
TehNut: This disallows creative players from creating a normal set of armor (eg: server admin recreating items a player lost).

This block and the one after could also be condensed since all that's changing is the amount.

src/main/java/WayofTime/bloodmagic/item/ItemBloodOrbTesting.java
```diff
@@ -0,0 +1,73 @@
...
+        setMaxStackSize(1);
+    }
+
+    @Override
+    public ActionResult<ItemStack> onItemRightClick(World world, EntityPlayer player, EnumHand hand) {
+        SoulNetwork network = NetworkHelper.getSoulNetwork(player);
+        if (!world.isRemote)
+            if (player.isSneaking()) {
+                network.setCurrentEssence(Integer.MAX_VALUE);
+                player.sendMessage(new TextComponentTranslation("commands.bloodmagic.network.fill.success", player.getDisplayName().getFormattedText()));
```
TehNut: `TextComponentTranslation` allows `ITextComponents` to be passed to it, so the user's name should remain as an `ITextComponent`.

src/main/java/WayofTime/bloodmagic/item/soul/ItemSoulGem.java
```diff
@@ -41,6 +41,12 @@ public ItemSoulGem() {
         setHasSubtypes(true);
         setMaxStackSize(1);
         setCreativeTab(BloodMagic.TAB_BM);
+        this.addPropertyOverride(new ResourceLocation("type"), new IItemPropertyGetter() {
+            @SideOnly(Side.CLIENT)
+            public float apply(ItemStack stack, World world, EntityLivingBase entityIn) {
+                return ((ItemSentientBow) RegistrarBloodMagicItems.SOUL_GEM).getCurrentType(stack).ordinal();
```
TehNut: ..?

TehNut (changes requested):

keraldi:
Done.
Searched for Will drain with regex and replaced manually.
Replaced Nausea with an additional check for creative activated MRS.

The creative (no) Will drain part is not sophisticated as it still requires Will to get the Will effects started, even if it doesn't drain them.

Issue #1396, opened 2018-08-08
[Suggestion] Implement ISentientSwordEffectProvider in sigils


ChrisThirtle:
ISentientSwordEffectProvider is an interface used to allow the off-hand item to provide bonus effects to the sentient sword (and also all of the sentient tools, but not the bow) when it hits a target. It has a basic (but turned off) implementation in the current Air Sigil code, but hasn't actually had any other effects coded, so I thought, why not put together that code.
I'm making this issue just so discussion about the effects, balance, and all that sort of thing can be put in one place and kept track of.

So, current ideas, and I'm allowing duplicates because there's only so many fitting things you can do with the current way the interface is implemented, basically just doing something either to the player or to a target on it getting hit:
 * Air Sigil: Levitation of course. Current duration is 10 seconds of levitation 1, I believe. That might be a bit much though.
 * Water Sigil: Drowning. Don't know how to make a target suddenly take drowning damage, but hey, there it is.
 * Lava Sigil: Light the target on fire. Best implementation would probably be making the sword act as if it had 1 level higher of Fire Aspect than it already does, otherwise it would conflict with said enchantment.
 * Void Sigil: A bit of extra void damage. Not a lot, but it ignores armor so that's pretty powerful.
 * Sigil of the Green Grove: Poison
 * Sigil of the Blood Lamp: Glowing
 * Sigil of Elemental Affinity: Negative fire resistance, so extra fire damage. Lorce also suggested making targets take extra water-related damage, but I've no idea how to do that.
 * Sigil of Magnetism: Negative jump boost. Pull them down to the ground. Extra falling damage and with high enough values can just plain stop things from jumping up blocks.
 * Sigil of Suppression: No idea.
 * Sigil of Haste: Slowness
 * Sigil of the Fast Miner: Mining Fatigue
 * Sigil of the Phantom Bridge: Nothing good here
 * Sigil of Compression: No idea.
 * Sigil of Ender Severance: Poisons endermen severely
 * Teleposition Sigil: Random teleport
 * Transposition Sigil: Random teleport, surprising I know
 * Sigil of the Claw: Weakness
 * Sigil of Elasticity: Increased knockback, perhaps?
 * Sigil of Winter's Breath: Slowness

Possible addition:
  * Make potion flasks implement the interface too and let those apply their effects to a target on hit. A use for all those negative potion effects, at least until we have throwable flasks.

keraldi:
[...]

PR #1569, opened 2019-03-26
ISentientSwordEffectProvider Implementation

Phil-M-H:
#1396 
Changed the interface to provide the will level too, so that the sigils could provide stronger effects depending on how much will is in the player's inventory.

Simplified the logic for weapons calling these methods. Instead of calling a separate to first check whether there was enough LP in the player's network, the applyEffectOnHit method will now return true if it applied the effect. 

Water and lava sigils provide drowning and lava damage, respectively, that increases with will level. LP drain also increases with will level.

Air sigil sends the target a few blocks up into the air, and increases the fallDistance counter relative to will level. 

Void sigil deals OUT_OF_WORLD damage to the player, but at half the amount of the water and lava sigils.

Elemental Affinity Sigil removes fire resistance and water breathing. If the target is a player, it will search their inventory and disable 1 elemental affinity sigil if the attacker has a will level < 3, or all of them if the attacker has a will level >= 3.

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/iface/ISentientSwordEffectProvider.java
+++ b/src/main/java/WayofTime/bloodmagic/iface/ISentientSwordEffectProvider.java
@@ -5,7 +5,17 @@ import net.minecraft.entity.EntityLivingBase;
 import net.minecraft.item.ItemStack;
 
 public interface ISentientSwordEffectProvider {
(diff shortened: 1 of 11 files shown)


keraldi:
?

Phil-M-H:
> ?

I don't think what I have is bad, but there it will probably be a while before I can do more and know how to do more. If you want I can reopen it though. I figured it wouldn't be nice to have this open PR for months.

PR #1571, opened 2019-03-28
Block range fixes

keraldi:
Fixed some block ranges that were still not reading from MRS.


src/main/java/WayofTime/bloodmagic/ritual/Ritual.java
```diff
@@ -216,6 +216,7 @@ public int getMaxHorizontalRadiusForRange(String range, List<EnumDemonWillType>
         return horizontalRangeMap.get(range);
     }
 
+    //TODO: Is this even needed? If yes, it should be fixed at some point (something similar should be in the RitualTinkerer).
```
TehNut: This comment is irrelevant to the PR

TehNut (changes requested):

Issue #1544, opened 2019-02-15
golem error


Shibva:
Hello Im aware that this is not following the template 
but I would like to notify you that after some version testing that Your Mod Blood Magic (2.4.0-102) is Causing an iterfearence to the Golems From Thaumcraft to Unabe to fly correctly due to recent changes.

please note that this wasent an issue until versions AFTER version 2.3,3-101
it was version 2.3.3-101 of the mod that the golems from Thaumcraft with there Flight upgrade were working correctly

Id hope that you can fix this issue 
I really like this mod lots

keraldi:
The template is there for a reason.
Unless you are not talking about just blood magic and still telling us version numbers and items you cannot expect us to know anything about what you're talking.

Fill out the issue template.
Provide the mod versions you tested it with which break the expected behavior (I assume it worked in BM-101 but broke in BM-102 with some Thaumcraft version (we need the thaumcraft version).

Verify that the issue exists even if you have only a minimum amount of mods (i.e. Thaumcraft and Bloodmagic and their dependencies).

Keep in mind that coremods can break stuff with a much higher probability than normal mods.

Give us the Thaumcraft version.
And for all sakes, provide a step-by-step guide to replicate the issue, that includes:

1. Names of all Items/mechanics used
2. how to click where, when and under which circumstances
3. ...

If you do not adhere to the issue template, the issue WILL be ignored, especially if it is an assumed or supposed cross-mod issue.

Shibva:
#### What happens: golems from Thaumcraft do Not Fly 



#### What you expected to happen: For golems to fly like the should of before i updated ONLY Blood Magic



#### Steps to reproduce:
Install 
[s-1.0.zip](https://github.com/WayofTime/BloodMagic/files/2871391/s-1.0.zip) via twich

2 load game 
3 make a world
4 Create a Golem with the arcrane levitator upgrade
5 Place Golem at any height (recommend 4 blocks or higher to see results)

if the golem moves and falls when it should be levitating then that is the problem

...

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: 2.4.0-102
- Minecraft: 1.12.2
- Forge: 14.23.5.2807

wrincewind:
There's a lot of mods in that modlist! have you tried disabling every mod aside from Blood Magic and Thaumcraft and trying this again? (In a new world, obviously)

Shibva:
Ive have determend this

[...]

PR #1576, opened 2019-04-05
Inter-mod gravity issue

Phil-M-H:
#1544 
Ritual of grounding logic for corrosive demon will was applying gravity to every entity on update.

Changed the logic so it would only apply no gravity to those that already have gravity, and would add gravity to entities that were on the noGravityList but are no longer have the suspended potion effect.


Iorce:
#1544

brunyman (commented):
This has to be included, fixes holograms crashing down instead of floating on multiplayer servers. Well done!

PR #1578, opened 2019-04-06
Veil of Evil & Ward of Sacrosanctity

keraldi:
No will stuff added yet.

Veil of Evil allows/forces spawns if they would happen under vanilla conditions, else enforce vanilla conditions if they are in a denied state.

Ward of Sacrosanctity disallows/denies spawns under vanilla conditions, else enforce vanilla conditions if they were allowed.


TehNut:
Fix conflict

Issue #1558, opened 2019-03-04
lava sigil stopped working


sujiko:
#### Issue Description:
lava sigil stopped putting lava source blocks onthe ground, but still puts down 50mb (1 bucket is 1000mb) of lava only into tanks. i reported it to the modpack author... which is just my friend and it still works in single player but works for no one on the server we have now. he isn't sure where to look if there is a way to fix it. If you cannot help with this, if you could guide us to where to perhaps look to fix it. we're thinking the amount for the bucket amount changed.

#### What happens:
the lava sigil only deposits a tiny bit of lava into a container and does not create source blocks now. there is more than enough LP(triple checked theres about 100,000LP) but will not create a lava source, the water sigil still works

#### What you expected to happen:
the lava sigil to put down a lava source block, or for it to deposit 1000mb(1 bucket) of lava into a tank.

#### Steps to reproduce:

1. really am not sure what caused it so i do not know the steps to reproduce. 
2. 
3. 
...

____
#### Affected Versions (Do *not* use "latest"):
we are using the .jar on a pack that has tinkers construct (the only tank i use it with) we also have tinkering with blood magic:
-TConstruct-1.12.2.12.0.115
-bloodtinker- 1.0.5
- BloodMagic: 1.12.2.4.0-102
- Minecraft: 1.12.2
- Forge: 14.23.5.2815

Phil-M-H:
Can you provide some actual steps to reproduce, or to setup the same mod-pack, so I can attempt to reproduce it?

alex2069:
I just got this bug as well (except for me it's 100mb).
Mod Pack: SevTech Ages v3.1.1
Blood Magic: BloodMagic-1.12.2-2.4.0-102
Tinker's Construct: TConstruct-1.12.2-2.12.0.115
Forge: 14.23.5.2815

It happened when I was filling up a Tinker's Seared Tank. The tank was not an even number a buckets (i.e. it was at 1900mb or 1100mb), and can store 4000mb.
I just kept right-click spamming to fill it up, when it was full, started doing it on the tank above it, but instead of doing a full 1000mb it only did 100mb. Thought it was a Tinker's bug, so right-clicked on ground but it didn't create a source block. Checked and had plenty of LP.

Give the Sigil to friend and he couldn't place source blocks either, and he could also only fill up the seared tanks with 100mb. Also tried creating a new Sigil but it too was behaving the same way.

[...]

PR #1585, opened 2019-04-22
UNTESTED: Possible fix for the FluidSigil issue

keraldi:
#1558

Lazy fix. Should probably find a better way than copying objects.


keraldi:
Archived. 1.12.2 Branches will not be maintained.

Keraldi:
.

Issue #1595, opened 2019-04-27
Performance issues sigils


runescapejon:
#### Issue Description:
Hello so, i have ran into an issue regarding of the sigils one of my one of players as using compression sigil to auto-compress stuff and this create performance issue. Here is a screenshot of the issue
![dfb5116cd06006f4e6ddac3e39476008](https://user-images.githubusercontent.com/10858509/56854995-5f8bce80-690d-11e9-87f3-3d0d0168ab9f.png)

#### What happens:
Performance issues from Compression sigils 

#### What you expected to happen:
some sort of fixed for this minor issue .-.

#### Steps to reproduce:

1. 
2. 
3. 
...

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: 1.12.2-2.4.0-102
- Minecraft: 1.12.2
- Forge: 14.23.5.2814

keraldi:
Any specifics on how much the player was compressing? It shouldn't be more inefficient than just crafting all of that manually, just in a shorter timeframe.

runescapejon:
it was actually 1 player that was doing it and he had mention that he was compressing a full inventory of items that about it

keraldi:
Consider me interested. Throw anything that you think could or should be optimized at me.

runescapejon:
What?

keraldi:
I meant that if you find other things that consume a lot of tick time, I'd like to see it.

I'd also like to see how that graphic looks like once a new build has been released.

TehNut:
I commented this on the PR, but the new improvements take my timings from about 15ms on average (15,000μs) down to 70μs on average.

keraldi:
@runescapejon new version with a fix is available, I would like to see the graph again, if you don't mind

runescapejon:
Thank you for fixing this, @Iorce 
![ee2c10a238ee749e189e7ce652d78746](https://user-images.githubusercontent.com/10858509/57985774-0bbb6380-7a3b-11e9-960c-8b34c9660d3d.png)

keraldi:
@runescapejon what's the monitoring tool you're using for this called?

TehNut:
https://github.com/sk89q/WarmRoast

runescapejon:
https://github.com/lucko/spark I use this but it's base on WarmRoast.

PR #1603, opened 2019-05-06
Improved Compression Sigil performance & fixed leftover eating bug

keraldi:
Compression reiterated!
- Fixed a bug that would consume compressable items even if the result could not be crafted because the minimum kept amount for that item has been reached

- added blacklist for uncompressable items
- added a map to quickly look up valid recipes if they have been crafted once already
- sped up the whole thing by adding break conditions


keraldi:
#1595

keraldi:
Done & tested thoroughly.

TehNut:
15ms on average -> 70μs on average

PR #1607, opened 2019-05-10
Add russian guide book

bigenergy:


TehNut:
Thanks for the contribution! Could you please remove any lines that are identical to the en_us lang? Since lang files are stacked, the en_us version will fall through to this one when in use and any changes we make up there will reflect down here.

bigenergy (commented):
Fixed

TehNut:
Sweet, thanks!

PR #1633, opened 2019-08-07
Fixes for Ritual of the Eternal Soul

sam-kirby:
* Use the same code to determine altar location and only cache its offset as is done in Well of Suffering. This prevents keeping a TileEntity loaded after it has been removed from the world, and also means the ritual will function if there are multiple master ritual stones using it in the world.
* Use getCapability to obtain the BloodAltar which implements IFluidHandler, and fill that instead of trying to fill the tile entity.
* Change the structure to match that found in previous versions.


keraldi:
Everything works except for setting the player health.

The main goal of setting the player health low is to prevent them from filling the altar manually.

In my opinion, just setting the health of the owner would not go far enough but setting the health of all players low would make it a weapon. A better solution would probably be to also prevent the altar from being filled from player sacrifices that are not from the owner.

keraldi:
I'll make one or two PRs to your fork to implement those functions incrementally.

keraldi:
I'm finished with the PR to your fork, you can merge it.

sam-kirby:
Done, thank you.

keraldi:
I should really thank you because my mind was apparently not in this world when I requested a merge for the unfinished and untested ritual in the first place >_<

Issue #1593, opened 2019-04-27
[suggestion] better automatable slates


kyrandita:
Infusing slates of the altar works just fine, and charging runes make it relatively quick to do by hand, but unless you build some redstone timer stuff with specific intervals on multiple altars for each of the tiers of slate, automation is tedious and ruins the aesthetic.
I envision some sort of special tank/smeltery that takes a blood orb corresponding to the tier of slate you want to create and it could goes straight from stone to the relevant level. consuming all the required sacrifice from the connected LP network of the orbs owner. maybe with some upgrade potential to reduce the total cost (more efficient infusion).
Another benefit to this later game system is that if the LP dries up in the middle of an infusion it doesn't lose progress anymore.
Compatibility with hoppers, only feeding out the finished slates, and redstone toggleable to pause production. Maybe combined with an LP network detector that provides a redstone signal when the connected LP network reaches a certain level (or drops below) could allow for good automation of rune and ritual materials.

Obviously you'd make this an efficiency increase instead of replacing the current method, maybe accessible once you reach a tier 3/4 altar meaning you still have to craft plenty of slates to get the various runes before you can even craft it.

Visually I like the idea of 4 arms suspending the blood orb above the stone, particles collecting into droplets of condensed life essence and dripping onto the stone below. But that's just an idea I had.

keraldi:
Network LP was never meant to achieve direct progress in crafting.

Automation is possible and in fact relatively easy with current methods, albeit the need for creative methods exists unless you utilize add-ons to Blood Magic.
Some weird I-can-do-it-all magical block is neither creative nor challenging.

The whole point of the Blood Altar is to have some sort of balancing that challenges a creative process. You can achieve everything in BM2 by just having a self-sacrifice altar and manual labor. If you want to achieve great things that are capable, you should try multi-altar setups with different methods of Life Essence acquisition. If you do it right, your altar should never run dry nor should the crafting take a long time.

You can even limit the crafted tier by temporarily removing a Blood Rune (with a Piston) and/or utilizing redstone timers.

I'm going to implement more features to make automation easier but I doubt that a single block will ever fully automate crafting for the Blood Altar in plain Blood Magic.

kyrandita:
[...]

Issue #1580, opened 2019-04-13
[Suggestion] Altar Functionality: Redstone Pulse


keraldi:
Altar could emit a redstone pulse when things finish crafting if a specific block is below the Altar (Redstone Lamp?).
This can be used for full altar automation with just Blood Magic.
The redstone pulse would prevent that the altar gets filled for a short time (hopper) and could activate a routing node for item extraction.

PR #1635, opened 2019-08-08
Implemented BloodAltar modification with redstone lamp below altar

keraldi:
- strong and weak Redstone signal of power 15 (similar to BlockLever)
- activates when crafting finishes and a BlockRedstoneLight is directly below the altar
- deactivates when the next crafting cycle is about to begin

Use case:

- Redstone-controlled automatic ejection of crafted goods

Tested functionality.
#1593 
#1580


src/main/java/WayofTime/bloodmagic/block/BlockAltar.java
```diff
@@ -180,4 +186,45 @@ public TileEntity createTileEntity(World world, IBlockState state) {
...
+    }
+
+    public int getMetaFromState(IBlockState state) {
+        int i = 0;
+
+        if (state.getValue(POWERED)) {
+            i = 1;
+        }
+
+        return i;
```
TehNut:
```java
return state.getValue(POWERED) ? 1 : 0;
```

src/main/java/WayofTime/bloodmagic/altar/BloodAltar.java
```diff
@@ -239,8 +242,14 @@ public void update() {
             tileAltar.getWorld().notifyBlockUpdate(tileAltar.getPos(), tileAltar.getWorld().getBlockState(tileAltar.getPos()), tileAltar.getWorld().getBlockState(tileAltar.getPos()), 3);
         }
 
-        if (internalCounter % 100 == 0 && (this.isActive || this.cooldownAfterCrafting <= 0))
+        if (internalCounter % 100 == 0 && (this.isActive || this.cooldownAfterCrafting <= 0)) {
+            /* Redstone Lamp below altar: Switch Off */
+            if (world.getBlockState(pos).getValue(BlockAltar.POWERED)) {
```
TehNut: Store this state lookup for later uses. Don't needlessly query the world multiple times.

TehNut (changes requested):

TehNut (approved):

PR #1636, opened 2019-08-11
Fixed Ritual Tinkerer & added range reset

keraldi:
Range is reset by right clicking (on the MRS) in "Define Area" Mode with a Blood Orb in the Offhand and the Ritual Tinkerer in the Main hand

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/item/ItemRitualReader.java
+++ b/src/main/java/WayofTime/bloodmagic/item/ItemRitualReader.java
@@ -2,6 +2,7 @@ package WayofTime.bloodmagic.item;
 
 import WayofTime.bloodmagic.BloodMagic;
 import WayofTime.bloodmagic.client.IVariantProvider;
+import WayofTime.bloodmagic.ritual.AreaDescriptor;
 import WayofTime.bloodmagic.ritual.EnumRitualReaderState;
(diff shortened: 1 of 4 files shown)


src/main/java/WayofTime/bloodmagic/item/ItemRitualReader.java
```diff
@@ -144,17 +154,21 @@ public EnumActionResult onItemUse(EntityPlayer player, World world, BlockPos pos
...
                                 IMasterRitualStone master = (IMasterRitualStone) tile;
                                 BlockPos pos2 = pos.subtract(masterPos);
                                 String range = this.getCurrentBlockRange(stack);
+                                if (range == null || range.isEmpty()) {
+                                    String newRange = master.getNextBlockRange(range);
+                                    range = newRange;
+                                    this.setCurrentBlockRange(stack, newRange);
+                                }
                                 Ritual ritual = master.getCurrentRitual();
                                 //TODO: Fix AreaDescriptor area handling to be inclusive, then remove the "-1" for range calculation below.
```
TehNut: If you're removing the `-1`, you should probably remove this comment as well.

TehNut (changes requested):

TehNut (approved):

PR #1638, opened 2019-08-11
Changed SoundType and visible Material of wooden path blocks from stone/rock to wood

keraldi:
Breaking wooden path blocks is now effective & faster with an axe and slow with a pickaxe type tool.

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/block/BlockPath.java
+++ b/src/main/java/WayofTime/bloodmagic/block/BlockPath.java
@@ -9,6 +9,7 @@ import net.minecraft.block.SoundType;
 import net.minecraft.block.material.Material;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.client.util.ITooltipFlag;
+import net.minecraft.entity.EntityLivingBase;
 import net.minecraft.item.ItemStack;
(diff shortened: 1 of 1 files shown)


src/main/java/WayofTime/bloodmagic/block/BlockPath.java
```diff
@@ -42,6 +43,22 @@ public void addInformation(ItemStack stack, World world, List<String> tooltip, I
         super.addInformation(stack, world, tooltip, tooltipFlag);
     }
 
+    @Override
+    public void onBlockPlacedBy(World worldIn, BlockPos pos, IBlockState state, EntityLivingBase placer, ItemStack stack) {
+        super.onBlockPlacedBy(worldIn, pos, state, placer, stack);
+        if (getMetaFromState(state) <= 1) {
+            setSoundType(SoundType.WOOD);
```
TehNut: That is not how this works.

TehNut (changes requested):

src/main/java/WayofTime/bloodmagic/block/BlockPath.java
keraldi: modifying the getter now.

Issue #1648, opened 2019-08-31
Altar can have negative progress


keraldi:
#### Issue Description:
![image](https://user-images.githubusercontent.com/23401986/64068508-901ba600-cc28-11e9-9316-c99e9eeba234.png)

#### What happens:
Altar can have negative progress (example with -2), probably at most negative drain rate.

#### What you expected to happen:
0 progress if all LP is drained

#### Steps to reproduce:

1. Build a T2 altar
2. Produce a not divisible by 5 amount of LP in the altar by using displacement runes and destroying them before the buffer is filled
3. Try to craft an item with a high drainage rate that is not a divisor of the current amount of essence in the altar and with an amount needed to craft that exceeds the currently present essence in the altar
...

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: -103 (compiled from source)
- Minecraft: 1.12.2
- Forge: -2808

PR #1649, opened 2019-08-31
Fixes BloodAltar being able to obtain negative progress.

keraldi:
#1648


src/main/java/WayofTime/bloodmagic/altar/BloodAltar.java
```diff
@@ -299,6 +299,9 @@ private void updateAltar() {
             } else if (!hasOperated && progress > 0) {
                 progress -= (int) (efficiencyMultiplier * drainRate);
 
+                if (progress < 0)
```
TehNut: `Math.max()`

TehNut (changes requested):

TehNut (approved):

PR #1652, opened 2019-09-04
Lava Crystals can now set things on fire!

keraldi:


src/main/java/WayofTime/bloodmagic/item/ItemLavaCrystal.java
```diff
@@ -73,6 +83,27 @@ public Binding getBinding(ItemStack stack) {
...
 
+    @Override
+    public EnumActionResult onItemUse(EntityPlayer player, World worldIn, BlockPos pos, EnumHand hand, EnumFacing facing, float hitX, float hitY, float hitZ) {
+        pos = pos.offset(facing);
+        ItemStack itemstack = player.getHeldItem(hand);
+
+        if (!player.canPlayerEdit(pos, facing, itemstack))
+            return EnumActionResult.FAIL;
+
+        else {
```
TehNut: Nesting this in an `else` is unnecessary since it'll never be reached if the `if` passes.

TehNut (changes requested):

TehNut (approved):

Issue #1662, opened 2019-10-10
crash when linking the lava crystal


Winddbourne:
https://paste.dimdev.org/uvipinirav.mccrash

#### What happens:

I right-click the lava crystal in creative mode and the game crashes, 

#### What you expected to happen:
I expect the lava crystal to be bound to me so I can test things in the mod pack I'm working on. 

#### Steps to reproduce:

1. enter creative mode
2. place down a blood altar and fill it with the creative sacrifical dagger
3. bind a blood orb to yourself and place it in the alter, make sure it is getting blood into your network
4. get a lava crystal and attempt to bind it to you. 
...

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: 1.12.2-2.4.-104
- Minecraft: 1.12.2
- Forge: 14.23.5.28.47

keraldi:
Try right-clicking into the air to bind the crystal first. It crashes because the crystal is not bound.

PR #1663, opened 2019-10-10
Fixes Lava Crystal causing a crash when attempting to place fire when it is unbound.

keraldi:
#1662


src/main/java/WayofTime/bloodmagic/item/ItemLavaCrystal.java
```diff
@@ -88,19 +88,23 @@ public EnumActionResult onItemUse(EntityPlayer player, World worldIn, BlockPos p
         pos = pos.offset(facing);
         ItemStack itemstack = player.getHeldItem(hand);
 
-        if (!player.canPlayerEdit(pos, facing, itemstack))
-            return EnumActionResult.FAIL;
+        Binding binding = getBinding(player.getHeldItem(hand));
+        if (binding != null) {
```
TehNut: Use an early escape rather than nesting everything

TehNut (changes requested):

Issue #1617, opened 2019-06-23
Possible conflict with Hated Mobs - PotionEventHandlers


Gamerbolts:
#### Issue Description:
This is my custom modpack. I was testing things and with the Hated Mobs mod went to spawn their mobs and got a crash https://pastebin.com/e9JkkPHz I posted it on their github and he said it looked to be a blood magic problem.

#### What happens:
Upon spawning mobs from hated mods, game crashes

#### What you expected to happen:
Mob to spawn and not crash

#### Steps to reproduce:

1. Install Blood Magic
2. Install Hated Mobs
3. Spawn Mob from hated mobs
... crash

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: 2.4.1-103
- Minecraft: 1.12.2
- Forge: 14.23.5.2836

keraldi:
Out of curiosity: did you somehow summon a mob in an unloaded world?

Gamerbolts:
Nope I took the spawn egg and just tried to spawn it right in front of me in the Overworld

keraldi:
I can provide you with a custom .jar if you're on the discord server.

Gamerbolts:
oh cool. um not yet but I can get on there,

PR #1668, opened 2019-10-21
Fixes some possible NPEs in PotionEventHandler

keraldi:
#1617

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/potion/PotionEventHandlers.java
+++ b/src/main/java/WayofTime/bloodmagic/potion/PotionEventHandlers.java
@@ -93,19 +93,23 @@ public class PotionEventHandlers {
 //            }
 //        }
         List<EntityLivingBase> noGravityList = noGravityListMap.getOrDefault(event.getEntityLiving().getEntityWorld(), Lists.newArrayList());
-        if (eventEntityLiving.isPotionActive(RegistrarBloodMagic.SUSPENDED) && !eventEntityLiving.hasNoGravity()) {
-            eventEntityLiving.setNoGravity(true);
-            noGravityList.add(eventEntityLiving);
-        } else if (noGravityList.contains(eventEntityLiving)) {
-            eventEntityLiving.setNoGravity(false);
-            noGravityList.remove(eventEntityLiving);
-        }
+        if (noGravityList != null)
+            if (eventEntityLiving.isPotionActive(RegistrarBloodMagic.SUSPENDED) && !eventEntityLiving.hasNoGravity()) {
+                eventEntityLiving.setNoGravity(true);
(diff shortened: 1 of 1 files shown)


src/main/java/WayofTime/bloodmagic/potion/PotionEventHandlers.java
```diff
@@ -93,19 +93,23 @@ public static void onEntityUpdate(LivingEvent.LivingUpdateEvent event) {
...
 //        }
         List<EntityLivingBase> noGravityList = noGravityListMap.getOrDefault(event.getEntityLiving().getEntityWorld(), Lists.newArrayList());
-        if (eventEntityLiving.isPotionActive(RegistrarBloodMagic.SUSPENDED) && !eventEntityLiving.hasNoGravity()) {
-            eventEntityLiving.setNoGravity(true);
-            noGravityList.add(eventEntityLiving);
-        } else if (noGravityList.contains(eventEntityLiving)) {
-            eventEntityLiving.setNoGravity(false);
-            noGravityList.remove(eventEntityLiving);
-        }
+        if (noGravityList != null)
```
JonesiBlitz: No opening bracket for this case.

keraldi: No opening bracket needed, as there is only one statement below it (another conditional).

Keraldi: No opening bracket needed, as there is only one statement below it (another conditional).

TehNut: A multi-line statement should still be wrapped for readability

src/main/java/WayofTime/bloodmagic/potion/PotionEventHandlers.java
```diff
@@ -93,19 +93,24 @@ public static void onEntityUpdate(LivingEvent.LivingUpdateEvent event) {
...
+                noGravityList.add(eventEntityLiving);
+            } else if (noGravityList.contains(eventEntityLiving)) {
+                eventEntityLiving.setNoGravity(false);
+                noGravityList.remove(eventEntityLiving);
+            }
+        }
+        
+        if (eventEntityLiving.isPotionActive(RegistrarBloodMagic.GROUNDED)) {
+            PotionEffect activeEffect = eventEntityLiving.getActivePotionEffect(RegistrarBloodMagic.GROUNDED);
+            if (activeEffect != null)
```
JonesiBlitz: Another multi line block is needed here.
While single line if statements don’t need wrapped it is almost easier to just have all if statements Wrapped for clarity. It is a team level decision though at most places.

PR #1669, opened 2019-10-21
Moved Boost into the class it belongs to & improved it a bit

keraldi:
Moved Boost into the class it belongs to.
Added a vertical flight amplification.
Changed the conditions to allow for negative boost.

TODO: jumping with boost still feels unnatural (too steep compared to sprint)
TODO: vertical movement seems only possible by directly modifying motionY which can escalate quickly

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/potion/PotionEventHandlers.java
+++ b/src/main/java/WayofTime/bloodmagic/potion/PotionEventHandlers.java
@@ -79,11 +79,8 @@ public class PotionEventHandlers {
         }
 
         if (event.getEntityLiving().isPotionActive(RegistrarBloodMagic.BOOST)) {
-            float percentIncrease = 0;
-            int i = event.getEntityLiving().getActivePotionEffect(RegistrarBloodMagic.BOOST).getAmplifier();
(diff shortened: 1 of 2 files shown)


src/main/java/WayofTime/bloodmagic/potion/PotionEventHandlers.java
```diff
@@ -77,21 +77,27 @@ public static void onEntityUpdate(LivingEvent.LivingUpdateEvent event) {
...
-//                }
-//            }
-//        }
+
+        if (event.getEntityLiving().isPotionActive(RegistrarBloodMagic.BOOST)) {
+            float percentIncrease = 0;
+            int i = event.getEntityLiving().getActivePotionEffect(RegistrarBloodMagic.BOOST).getAmplifier();
+            {
+                percentIncrease += (i + 1) * 0.5F;
+            }
```
JonesiBlitz: Minor suggestions:
- The float could just be declared on line 85 since there is only a single calculation.
- The int on line 83 could use a better name than I, will make future reference easier to see what the value is just referring to the name.

Issue #1625, opened 2019-07-24
#Box in Input Routing Nodes effectiveness/utility


VincentLongiug:
#### Issue Description:

As far as I can see, the #Box in the Input Nodes forces a certain quantity of items to stay in the selected inventory but...

#### What happens:

When I made a simple Routing system to test how it works, I noticed that this #Box don't work properly: if you have Cobblestone (# 20) and Iron Ore (# 10) in the Filter (for instance), you will either have 20 Cobblestones remaining, or 10 Iron Ores, or none of them if you add an item which must stay in the chest (i.e not set in the Filter) (more details in "Steps to reproduce").

#### What you expected to happen:

If that #Box in the Input Nodes is supposed to have the purpose I stated above, there should always remain in this chest 20 Cobblestones, 10 Iron Ores and all the other Blocks/Items I don't want to transfer.
Else, the #Box in the Input Nodes should be removed (or converted into an Item priority).
NB: The #Box of the Output Nodes works perfectly as expected, for what I've seen.

#### Steps to reproduce

Here is one of the various methods that illustrate the problem :
1. Create a simple Routing system (from N to S : Chest 1, Input Node, Master Node, Output Node, Chest 2).
2. Put a Precise Item Filter, a Cobblestone (# 20) and an Iron Ore (# 10) in the Input Node (section N).
3. Put 64 Iron Ores in Chest 1, it will remain 10 Iron Ores.
4. Put 64 Cobblestone in Chest 1 on the left of the remaining Iron Ores, it will remain 20 Cobblestones and no Iron Ores (if you put the Cobblestones on the right of the Iron Ores, only 10 Cobblestones and no Iron Ore will remain).
5. Put 20 Seeds 1 by 1 in Chest 1, you will see the the remaining Cobblestones be transferred 1 by 1.
6. While the 20 Seeds are still in place, putting any Cobblestone or Iron Ore in Chest 1 will cause their transfer (the Seeds won't move).

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: 2.4.1-103
- Minecraft: 1.12.2
- Forge: 14.23.5.2838

*Note of the author: even if I may use Filters on Input Nodes, I only use #Boxes in Outputs, so I don't care if that inconvenience is fixed or not, it's just... When I test ! OH MAN ALIVE !!!*

wrincewind:
Closing as stale - 1.12 is no longer supported. Sorry!

PR #1678, opened 2020-01-18
Various routing node improvements/fixes

### LINKED ISSUE #1657: Fluid routing issues
#### Issue Description:

Hello,

I have encountered a problem when using Routing Nodes to transfer fluids. Specifically, the system does not handle multiple inputs and outputs within the same routing network correctly.

This is a minimal problematic setup:
Create a routing network with one Master Routing Node, one Input Routing Node, and two Output Routing Nodes. Place a fluid tank next to each of the I/O nodes. Place a Fluid Filter in each of the I/O nodes, configured in the right direction and with the same fluid.

Test 1:
Start adding fluid to the "input" tank, one bucket at a time.

Expected behavior:
Well ideally I would love for the fluid to be split evenly between both output tanks; although I can see that the system is not set up to handle this well.
Either way, I would expect that after adding enough fluid to the input tank, both output tanks will be filled.

Actual behavior:
Only the first output tank ("first" in priority order) ever receives any fluid. Even after the first output tank is full, the second tank never receives any fluid at all, even if there is enough fluid in the input tank.

Test 2:
Add multiple buckets of fluid to the input tank at once (right-clicking with another filled tank, or using a transfer method that can push more than 1000 mB at once.

Expected behavior:
The network transports a maximum of 1000 mB per operation.

Actual behavior:
The network transports up to 1000 mB per operation _per connection_. In the two outputs case, both output tanks receive 1000 mB of fluid with every operation.

#### Where the problem is:

I have looked at the source, and I believe I have figured out the issue:

https://github.com/WayofTime/BloodMagic/blob/1caae69992618c1b7a5bba60a79c4ea664717a91/src/main/java/WayofTime/bloodmagic/routing/RoutingFluidFilter.java#L151

The `transferThroughInputFilter` method returns the amount of fluid _remaining_ in the fluid stack being transferred; while the call of this method in

https://github.com/WayofTime/BloodMagic/blob/1caae69992618c1b7a5bba60a79c4ea664717a91/src/main/java/WayofTime/bloodmagic/tile/routing/TileMasterRoutingNode.java#L159

expects to get the amount of fluid that was actually _transferred_.

Therefore, if the transfer of fluid to the first output fails (for example, because the output tank is full, or the output filter is set to accept a different type of fluid), `RoutingFluidFilter.transferThroughInputFilter` returns 1000 (the maximum transfer amount), which then `TileMasterRoutingNode` reads as "everything was transferred, we are done", and it stops looking for other possible outputs.

This makes setting up any non-trivial network (basically anything other than single input - single output) impossible; as when _one_ transfer fails, _all other_ transfers are abandoned.

Conversely, if the entire fluid stack is successfully transferred, `transferThroughInputFilter` returns 0, and `TileMasterRoutingNode` continues looking for more connections, possibly exceeding the maximum transfer rate.

Note that the item transfer code is correct; it returns the number of items that were transferred. The problem is only with the transfer of fluids (or rather, the inconsistency between the two -- I believe the item transfer is working as intended).

#### Proposed fix:

Change

https://github.com/WayofTime/BloodMagic/blob/1caae69992618c1b7a5bba60a79c4ea664717a91/src/main/java/WayofTime/bloodmagic/routing/RoutingFluidFilter.java#L151

to

    return drained;

This fixes the expected return value of that function.

#### Affected Versions

Testing was done in the FTB Interactions modpack:
- BloodMagic: 1.12.2-2.4.1-103
- Minecraft: 1.12.2
- Forge: 14.23.5.2838

However the issue is present in the current source code (branch 1.12) as it is now on github. (See above.)

michaelPoul:
Hi, here are 3 routing nodes bug fixes and one performance optimization:

- Make Fluid output filters respect allowedAmount
Fluid output filters would try to extract the amount given to them by the input filter, ignoring their own limits. This caused odd behavior where inputs limited to amounts such as 50 mb would insert 1000mb in the tank.
- Fix input item routing filters not properly filtering
All the input item filters derived from PreciseItemFilter had odd behavior since empty item stacks are treated like air in many comparisons.
- Fix fluid routing with multiple nodes (fix #1657)
see related issue
- Improve item routing perf by reducing ItemStack copies
Changes the item filter code to make less ItemStack copies and do less simulated ItemStack extractions, improving performance in large routing systems. Also adds a canSkip heuristic to prune filters that won't be used.
- Rename TestItemFilter -> PreciseItemFilter
I know the contributing guidelines say not to do unnecessary file renames, but I feel this one is useful since it better matches the ingame item name. Also, I think the old name was meant to be temporary ;). I can remove this commit though.

The commit messages have more information as well! I tested the routing out on a sample world w/one input and one output of each type (including default filters) and several tanks. It seems to work

### DIFF
--- a/src/main/java/WayofTime/bloodmagic/item/routing/ItemRouterFilter.java
+++ b/src/main/java/WayofTime/bloodmagic/item/routing/ItemRouterFilter.java
@@ -58,11 +58,11 @@ public class ItemRouterFilter extends Item implements IItemFilterProvider, IVari
 
     @Override
     public IItemFilter getInputItemFilter(ItemStack filterStack, TileEntity tile, IItemHandler handler) {
-        IItemFilter testFilter = new TestItemFilter();
+        IItemFilter testFilter = new PreciseItemFilter();
 
         switch (filterStack.getMetadata()) {
             case 0:
-                testFilter = new TestItemFilter();
+                testFilter = new PreciseItemFilter();
                 break;
             case 1:
                 testFilter = new IgnoreNBTItemFilter();
@@ -101,7 +101,7 @@ public class ItemRouterFilter extends Item implements IItemFilterProvider, IVari
 
         switch (filterStack.getMetadata()) {
             case 0:
-                testFilter = new TestItemFilter();
+                testFilter = new PreciseItemFilter();
                 break;
             case 1:
                 testFilter = new IgnoreNBTItemFilter();
--- a/src/main/java/WayofTime/bloodmagic/routing/DefaultItemFilter.java
+++ b/src/main/java/WayofTime/bloodmagic/routing/DefaultItemFilter.java
@@ -37,81 +37,49 @@ public class DefaultItemFilter implements IItemFilter {
     }
 
     /**
-     * This method is only called when the output inventory this filter is
-     * managing receives an ItemStack. Should only really be called by the Input
-     * filter via it's transfer method.
-     *
-     * @param inputStack - The stack to transfer
-     * @return - The remainder of the stack after it has been absorbed into the
-     * inventory.
-     */
-    @Override
-    public ItemStack transferStackThroughOutputFilter(ItemStack inputStack) {
-        int allowedAmount = inputStack.getCount(); //This is done to make the migration to a maximum amount transfered a lot easier
-
-        if (allowedAmount <= 0) {
-            return inputStack;
-        }
-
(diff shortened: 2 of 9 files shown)


michaelPoul:
Looks like 0823493 was an already known issue: https://github.com/WayofTime/BloodMagic/issues/1625

keraldi (approved):

PR #1713, opened 2020-11-19
Creating a usable API

Arcaratus:
- Moved ~~a lot of~~ some things/interfaces over to the API ~~(altar, incense, orb, will, some interfaces)~~
  ~~- **Binding** and **SoulTicket** have also been moved over~~
- Destroyed each and every internal reference in the API so that it's actually usable
- Implement a dummy API instance for other people to use properly
- Added a few comments to explain what everything does


WayofTime:
TehNut's statute of limitations has elapsed.

PR #1732, opened 2020-12-27
Fix non [a-z0-9/._-] character in path of location

Victorsueca:
Fixes a net.minecraft.util.ResourceLocationException that happens when the mod is loaded with Actually Additions due to the third party crop integration using an old block ID with deprecated casing.


WayofTime:
Thanks for the PR!

PR #1737, opened 2020-12-30
Even Better Functional Array Template!

VT-14:
Moved Functional Array Output Texture Up 1 Pixel

To fix the alignment issue.  Hey, the only example I had at the time had a protrusion on the top!


WayofTime:
Fair, fair!

PR #1759, opened 2021-01-22
Update nyi.json

wrincewind:
We've added a few rituals since i wrote this. Oops. 
[note: double check that it's right? Please?]


WayofTime:
Ritual of Regeneration is in.

PR #1772, opened 2021-01-27
Make tooltips gray

Arcaratus:
Made the tooltips on blocks and items all gray
Also idk what happened to the git history ¯\_(ツ)_/¯


WayofTime:
:V

PR #1788, opened 2021-02-20
Added Living Armor Upgrade Tables

VT-14:
These Tables are made using a processor, so they are accurate for the current instance.

The text can be translated using the language files.


VT-14:
I've now also added Localization support to the ARC Processor, and made the following tweaks.

1. Fixed leading whitespace in Well of Suffering's multiblock, so it will line up properly on GitHub.
2. Corrected Extra Recipe Mapping for the Controlled Charge.  This removed an error from the log and properly links the item.
3. Re-added Extra Recipe Mappings for Living Armor.  I suspect a change in Patchouli lets this work properly now.

PR #1834, opened 2021-08-03
i18n for the guide book

TUsama:
Due to the pachoulii load method, the guide book cannot be translated unless you use the i18n key. So I patch it.


VT-14:
This is similar to #1798, but I think that got ignored because WayofTime was busy.

`en_us.json` is made using a Data Generator.  In order for the language file change to stay around, the changes also have to be made in `src/main/java/wayoftime/bloodmagic/common/data/GeneratorLanguage.java`.

Notes for potential translators:

1) The "patchouli.bloodmagic.(etc)" translation keys are planned to be moved to "guide.patchouli.bloodmagic.(etc)" keys in the near future.  Any portion of an Entry generated via code should work with translations set there.

2) Not using the language files for the actual Patchouli Entries (book.json's `i18n` option, default false) is an intentional decision.  It would make working with the book far more tedious (reload all resources every time you change an entry vs. just reloading the book), and changing language will almost certainly lengthen or contract text to the point where page formatting would have to change.  Patchouli's default method of translations is to have translated versions of the book in their own folders (same level as the en_us folder).  The en_us version is the master, and translations can replace those entries.

TUsama:
Do you mean that I should add the keys at [there](https://github.com/WayofTime/BloodMagic/blob/e84e9d7fb10a2d919856c1bc1a176f04e63d0c3d/src/main/java/wayoftime/bloodmagic/common/data/GeneratorLanguage.java#L378)?

VT-14:
Yup.

Don't worry about the patchouli.bloodmagic ones yet.  I'll change those sometime after these changes are pulled.

src/main/resources/data/bloodmagic/patchouli_books/guide/book.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "guide.bloodmagic.name",
-  "subtitle": "Alchemical Wizardry",
-  "landing_text": "Welcome to $(6)Blood Magic$()! $(br2)$(l:bloodmagic:utility/nyi)A lot of stuff$() isn't yet implemented, so please excuse our dust. $(br2)Click $(l:bloodmagic:utility/getting_started)HERE$() to get started. If you find any bugs, please report them on our $(l:https://github.com/WayofTime/BloodMagic/issues)Github$().",
+  "subtitle": "guide.bloodmagic.name",
```
VT-14: I think this is supposed to be guide.bloodmagic.subtitle.

TUsama:
sorry for that😓

VT-14 (approved):

WayofTime:
👍

PR #1841, opened 2021-08-30
Feature Redstone

damned-me:
# Redstone Implementation
Implemented behavior of the altar to the redstone, as reported in the manual.

According to the block below the altar
| Block| Effect | 
| -------- | ---------- | 
| Default  | Comparator Reads LP | 
| Bloodstone| Comparator reads network LP of the Player's Orb inside the altar | 
| RedstoneLamp | Triggers a softPower (15) 1 tick signal when a crafting operation ends |

## Changes
- The Block checks the current redstone mode, Default and Bloodstone mode are managed in the `BloodAltar.java` class, RedstoneLamp in `TileBloodAltar.java`

- Moved Properties of Bloodstone (for brick and stone)  from `BloodMagicBlocks.java` to `BloodstoneBlock.java` as a standalone class, now used to check if the block below the altar is Bloodstone

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/altar/BloodAltar.java
+++ b/src/main/java/wayoftime/bloodmagic/altar/BloodAltar.java
@@ -3,6 +3,7 @@ package wayoftime.bloodmagic.altar;
 import com.google.common.base.Enums;
 
 import net.minecraft.block.BlockState;
+import net.minecraft.block.RedstoneLampBlock;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.CompoundNBT;
 import net.minecraft.particles.ParticleTypes;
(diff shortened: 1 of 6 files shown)


src/main/java/wayoftime/bloodmagic/altar/BloodAltar.java
```diff
@@ -756,6 +764,32 @@ public AltarTier getCurrentTierDisplayed()
 		return currentTierDisplayed;
 	}
 
+	public int getAnalogSignalStrenght(int redstone_mode)
```
WayofTime, line 767: The parameter name should be camelCase to match current format

WayofTime (dismissed):
Unless TehNut has anything else to say about it, I'll need to give it a test to make sure it works. Thanks for the submission!

WayofTime:
Works well for all three stated cases.

PR #1845, opened 2021-09-08
Add Elytra Upgrade for Living Armor

VT-14:
* Upgrade Tome crafted by using Arcane Ash to combine Binding Reagent and an Elytra.  Due to confusion with the data generator, this recipe is currently made by hand (not generated).
* The upgrade matches vanilla Elytra mechanics, though has the unique base texture from 1.12.
* When in use, durability is drained from the Chestplate, but at half the rate the vanilla Elytra does (1 point every 40 ticks rather than 20).
* There is only a single level for this upgrade, and it costs 15 Upgrade Points.
* Documentation in Patchouli is included, though Wrince may want to review it before release.

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/BloodMagic.java
+++ b/src/main/java/wayoftime/bloodmagic/BloodMagic.java
@@ -1,11 +1,15 @@
 package wayoftime.bloodmagic;
 
+import java.util.Map;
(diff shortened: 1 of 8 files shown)


WayofTime:
Does what it says on the tin!

Issue #1838, opened 2021-08-13
Tartaric Gems with aspects don't work in the forge


TwistedPorkchop:
Like the title says, when you have a tartaric gem with a demon will aspect associated (I only tried corrosive) it won't craft anything with the forge.

PR #1847, opened 2021-09-09
Allow the Soul Forge to use any type of Will for Recipes.

VT-14:
#1838

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/tile/TileSoulForge.java
+++ b/src/main/java/wayoftime/bloodmagic/tile/TileSoulForge.java
@@ -135,7 +135,7 @@ public class TileSoulForge extends TileInventory implements ITickableTileEntity,
 		EnumDemonWillType typeInGem = EnumDemonWillType.DEFAULT;
 		for (EnumDemonWillType type : EnumDemonWillType.values())
 		{
-			soulsInGem += getWill(type);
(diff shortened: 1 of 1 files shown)


src/main/java/wayoftime/bloodmagic/tile/TileSoulForge.java
```diff
@@ -131,7 +131,17 @@ public void tick()
 			}
 		}
 
-		double soulsInGem = getWill(EnumDemonWillType.DEFAULT);
+		double soulsInGem = 0;
+		EnumDemonWillType typeInGem = EnumDemonWillType.DEFAULT;
+		for (EnumDemonWillType type : EnumDemonWillType.values())
+		{
+			soulsInGem = getWill(type);
+			if (soulsInGem > 0)
```
WayofTime: Should perform the check versus the minimum amount of will in the recipe, in the case of there being a gem with multiple aspects.

wrincewind: A gem with multiple aspects? I'd never even considered that as a possibility... But it sounds like it'd be pretty cool.

VT-14: Adjusted by the latest commit.

Issue #1855, opened 2021-10-08
Stacia Expert Bloodmagic based crash


Olorinmaiar:
#### Issue Description:
Note: If this bug occurs in a modpack, please report this to the modpack author. Otherwise, delete this line and add your description here. If this is a feature request, this template does not apply to you. Just delete everything.

#### Steps to reproduce:

1. I tried to re-join
2. tried to copy the world

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: 1.16.4-3.1.4-24
- Minecraft: 1.16.5
- Forge: forge-36.2
[crash-2021-10-08_22.04.00-server.txt](https://github.com/WayofTime/BloodMagic/files/7313931/crash-2021-10-08_22.04.00-server.txt)

wrincewind:
looks to be related to #1853, they're both crashing due to the elytra upgrade, at least according to the first line of both crashlogs.

Olorinmaiar:
And what can I do to fix this problem?

wrincewind:
Currently, i'd advise making a backup of your world, then trying to downgrade Blood Magic to the previous version and see if that helps. If it does, take off the living armour and stick it in a chest, then you can upgrade again - or wait, if you'd prefer. 

We're looking into a more permanent fix right now.

Tristan-LeBras:
In my case I didn't do any upgrade (just to be sure we are talking about the same thing)

Olorinmaiar:
How do i Downgrade the Blood Magic Mod?

wrincewind:
https://www.curseforge.com/minecraft/mc-mods/blood-magic/files/3475176

Download this version (it's the second-newest one, one version behind) and replace the version you have on all clients and servers with this one.

keraldi:
https://github.com/WayofTime/BloodMagic/pull/1867

wrincewind:
Closing as 1.16 is out-of-support.

Issue #1853, opened 2021-10-06
[1.16.5] Client-crash after putting living armor


Tristan-LeBras:
#### Issue Description:
Client-crash after putting the 'living armor' in Valhesia 3.4.4 on a server (linked the crash report to Val3 team and they asked me to put it here).
Cannot relog 'internal server error' .

#### What happens:
I just crafted the full set of 'living armor' and the moment I put it on I crashed.
After that I'm not able to connect to the server 'internal server error' but others can and they can play normally.

#### Steps to reproduce:
Don't know if Valhesia 3 has an impact here (didn't test the mod alone)
1.  Put the living armor
2.  Enjoy the crash
3. 
...

____
#### Affected Versions (Do *not* use "latest"):
Valhesia 3.4.4 launched via Curseforge
- BloodMagic: BloodMagic-1.16.4-3.1.5-25
- Minecraft: 1.16.5
- Forge: 36.2.8

Link to log + crash report :
https://gist.github.com/Tristan-LeBras/188c165e08d8e117d43edce40c22d1a3

the-first-ideal:
Tried the downgrade, that didn't work, so as admin of the server I spammed /clear [playerName] bloodmagic:livingplate over and over as my player joined the server. It finally caught him, so now he can play. 
It does seem like though that there is a potential fix from user VT-14, just needs to be implemented.

Takeuout:
Putting on the full set of living armor completely breaks the game... had to roll back the server.
i have seen posts about this going back as far as 2017 does this actually ever plan on getting fixed?

enigmatica 6 
Minecraft Version: 1.16.5
Modpack Version: 0.5.20

[05.11 16:54:22] [Server] Caused byjava.lang.NullPointerException: Cannot invoke "wayoftime.bloodmagic.core.living.LivingStats.getLevel(net.minecraft.util.ResourceLocation)" because the return value of "wayoftime.bloodmagic.core.living.LivingStats.fromPlayer(net.minecraft.entity.player.PlayerEntity)" is null
[05.11 16:54:22] [Server] 	at wayoftime.bloodmagic.common.item.ItemLivingArmor.hasElytraUpgrade(ItemLivingArmor.java:224) ~[bloodmagic:1.16.4-3.1.5-{re:classloading,pl:runtimedistcleaner:A}
[...]

PR #1856, opened 2021-10-09
Fix Blood Elytra from crashing if the Living Armor Stats aren't…

VT-14:
…Initialized when hasElytraUpgrade is checked

Also slightly streamlined the checks.

This should fix the issue with Caelus checking canElytraFly as soon as a fresh Living Armor Chestplate is equipped.

#1855, #1853


HighPriestWa:
Im pretty new to this whole mod thing so I'm kind of lost. I understand that I need to edit the file in order to solve the living armor issue.

The problem I'm having is I'm hosting the modded server off the ggservers website and I'm struggling to find the file you are referencing. I understand that its in the mods root folder but I don't know where to find that folder.

Any help finding the file would be greatly appreciated,

WayofTime:
Used the fix in #1867 instead.

PR #1891, opened 2022-05-24
Non-Data Gen Brilliance Upgrade Recipe

VT-14:
The NBT wasn't getting set correctly, and I'm not sure how to fix it in the data generator files.

This is the method I used for the Elytra and Curios Slot upgrades previously.


WayofTime:
Fixed properly now in latest commit. Thanks!

Issue #1904, opened 2022-07-06
Spectral Block doesn't render and crashed when placed (Large modpacks)


Ryuushen:
I saw that Spectral Block doens't render and so i wanna try to see if its functional at least but when i placed it on the floor it crashed the game

- BloodMagic: 3.1.12-38
- Minecraft: 1.16.5
- Forge: 36.2.35
[test.txt](https://github.com/WayofTime/BloodMagic/files/9051202/test.txt)

Ryuushen:
[crash-2022-07-05_20.15.02-server.txt](https://github.com/WayofTime/BloodMagic/files/9051204/crash-2022-07-05_20.15.02-server.txt)
The Test.txt up there is wrong file. This is the crash log

WayofTime:
Thanks for the report. The Spectral Blocks are not something that will normally be placed by the player - as such, I didn't include a few null checks to see if it actually contains a fluid. I'll remove it from the Creative tab, and maybe add null checks to make sure this doesn't occur.

wrincewind:
closing as 1.16 is no longer supported. (fixed in 1.20)

PR #1911, opened 2022-07-21
Fix TileSpectral

PackmanDude:
#1904


TehNut:
This is not a valid fix for the problem.

PackmanDude:
> This is not a valid fix for the problem.

Thanks, I noticed it already. It keeps returning FATAL-errors, because of the Spectral block. Dev said 2 weeks ago he would fix that with some tests and other things 🙁

WayofTime:
If you're referring to https://github.com/WayofTime/BloodMagic/issues/1904 , I said that I'd remove it from the creative tab. I didn't say for sure that I would alter the tile entity itself.

PackmanDude:
Okay.
1.16.5 was released ≈5 years ago and there are still common bugs 😔

WayofTime:
You blow my mind.

PR #1926, opened 2022-08-28
Added Gilded Living Armour Upgrade

VT-14:
Gilded makes the Piglins found across the Nether passive, as if you were wearing a piece of Gold Armor.

Without this upgrade, Piglins are hostile, and appeasing them by wearing a piece of Gold Armor means you aren't wearing a full set of Living Armour, so you don't get the buffs.

Current Stats: Costs 5 Upgrade Points.  Crafted using a Binding Array and a Block of Gold.

This includes documentation in the Sanguine Scientiem.


src/main/java/wayoftime/bloodmagic/common/item/ItemLivingArmor.java
```diff
@@ -225,4 +225,13 @@ public boolean hasElytraUpgrade(ItemStack stack, LivingEntity entity)
...
 			return false;
 	}
+
+	@Override
+	public boolean makesPiglinsNeutral(ItemStack stack, LivingEntity entity)
+	{
+		if (stack.getItem() instanceof ItemLivingArmor && entity instanceof Player && LivingUtil.hasFullSet((Player) entity))
+			return LivingStats.fromPlayer((Player) entity, true).getLevel(LivingArmorRegistrar.UPGRADE_GILDED.get().getKey()) > 0;
+		else
+			return false;
```
keraldi, line 235: Redundant conditional branch at the end of function (remove `else return false` if a function/method closes afterwards)

src/main/java/wayoftime/bloodmagic/common/item/ItemLivingArmor.java
```diff
@@ -225,4 +225,13 @@ public boolean hasElytraUpgrade(ItemStack stack, LivingEntity entity)
 		else
 			return false;
 	}
+
+	@Override
```
keraldi, line 229: Fairly sure that the `@Override` decorator is wrong here. If `makePiglinsNeutral()` does not exist in any super/parent class (not inherited), remove this line

keraldi (changes requested):

VT-14:
Closing this in favor of #1968 (updated for 1.20).

PR #1936, opened 2022-10-20
Added the ability to dye Living Armor

dphaldes:
Dyes can be used in a crafting table with the chest-plate to dye it. The armor color is only dependent on the color stored on the chest-plate.
Includes all 16 colors as separate textures to maintain detail for every color.
Thanks to ArsNouveau developers for help and Code References.

**TODO**
- [ ] Re-organise files+code 
- [x] Display color in the chest-plate tooltip
- [ ] Take care of Elytra trait
- [ ] Finalize textures
- [x] Test to retain traits and enchants.


src/main/java/wayoftime/bloodmagic/common/item/ItemLivingArmor.java
```diff
@@ -44,14 +41,23 @@ public ItemLivingArmor(EquipmentSlot slot)
 	@Override
 	public String getArmorTexture(ItemStack stack, Entity entity, EquipmentSlot slot, String type)
 	{
+		String color = "red";
+
+		for (ItemStack itemStack : entity.getArmorSlots()){
```
TehNut, line 46: The chestplate would always be found in the second slot, no? Is this loop necessary?

dphaldes: I don't know if there are ways to get the chest slot directly. I will try to replace it after researching a bit

TehNut: I haven't opened a Minecraft project in a while, but doesn't `getArmorSlots()` return either an array or a list? Slots don't just change position so you should just be able to get it by index.

dphaldes: That makes sense as I did encounter Empty ItemStacks for non-filled armorSlots. Didn't realize I could just get the specific index. I will make this change if it works the same for Itemstands as well

dphaldes: `getArmorSlots()` returns an `Iterable` not a list. Can't think of simple way to access the list directly

PR #1953, opened 2023-01-27
Deleted Mixin to fix Reference Map WARN

MrLoLf:
I deleted the Mixin file, because it wasn't needed to call any file and was throwing this WARN:
```
[main/WARN]: Reference map 'bloodmagic.refmap.json' for bloodmagic.mixins.json could not be read. If this is a development environment you can ignore this message
```


dphaldes:
#2005 adds mixins again ( 1.18.2 backport soon) and thus the refmap and mixins.json are required now

Issue #1919, opened 2022-08-14
[1.18] Mystical Agriculture integration failing


Chaos02:
#### Issue Description:
Am Modpack author. Log says Mystical agriculture integration fails.
```
[Worker-Main-5/ERROR] [bloodmagic/]: MysticalAgriculture integration cancelled: unable to find a class: com.blakebr0.mysticalagriculture.api.crop.ICrop
```

____

- BloodMagic: 1.18.2-3.2.5-40
- Minecraft: 1.18.2
- Forge: forge-40.1.73

WayofTime:
Thanks, will look at the changes to their methods later.

andrewsf:
I noticed that "[1.18 and older is NO LONGER SUPPORTED](https://github.com/BlakeBr0/MysticalAgriculture/issues/266)" but then breaking API changes were made in the 1.18 branch..

This MysticalAgriculture commit [b77e79cb138e9b8461c33b9489b3539a36aeb516](https://github.com/BlakeBr0/MysticalAgriculture/commit/b77e79cb138e9b8461c33b9489b3539a36aeb516) removed the interface `ICrop`.

Removing an interface is a little bit crappy to do in an "api" package. But it was marked:

```
// TODO: 1.17: remove
@Deprecated
```

https://github.com/WayofTime/BloodMagic/blob/69b627090c55910be40c69ed9f8bd58de5863e07/src/main/java/wayoftime/bloodmagic/ritual/harvest/HarvestHandlerPlantable.java#L175

Now you just use the `Crop` class directly. And I don't know if it matters here but `getCrop` returns a `net.minecraft.world.level.block.CropBlock`, not a `net.minecraft.block.CropsBlock` (class renamed from 1.16 -> 1.18).

wrincewind:
Closing as 1.18 is no longer supported. (resolved in 1.20+)

PR #1955, opened 2023-01-30
Mystical agriculture integration 

BrodyWills:
updated integration to work with new mystical agriculture API
#1919


BrodyWills:
#1919

PR #1961, opened 2023-08-05
added curios support for divination/seer sigil

stellanera98:
Changed `ElementDivinedInformation.shouldRender()` to consider Curios slots if the mod is loaded (`BloodMagic.curiosLoaded`) as well as the main and off hand

Changed the check for the sigils in above function a bit to avoid basically duplicating what's inside the if/else clause

Added the divination and seer sigil to the charm and living armour socket curios tags

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/client/hud/element/ElementDivinedInformation.java
+++ b/src/main/java/wayoftime/bloodmagic/client/hud/element/ElementDivinedInformation.java
@@ -3,10 +3,14 @@ package wayoftime.bloodmagic.client.hud.element;
 import java.util.List;
 
 import net.minecraft.client.Minecraft;
+import net.minecraft.core.NonNullList;
+import net.minecraft.world.InteractionHand;
 import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.level.block.entity.BlockEntity;
-import net.minecraft.world.InteractionHand;
+import net.minecraftforge.items.IItemHandler;
+import top.theillusivec4.curios.api.CuriosApi;
+import wayoftime.bloodmagic.BloodMagic;
 import wayoftime.bloodmagic.common.item.BloodMagicItems;
 import wayoftime.bloodmagic.common.item.sigil.ItemSigilHolding;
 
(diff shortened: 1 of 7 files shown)


VT-14:
When I initially added Curios compatibility I reworked BM's inventory handler.  It's set up to allow add-ons to be able to add their custom inventories via the API too.

I would also like to keep as much Curios related code as possible consolidate within https://github.com/WayofTime/BloodMagic/blob/1.18.2/src/main/java/wayoftime/bloodmagic/compat/CuriosCompat.java.  The current PR ends up with a bit of duplicated code already present in that class.

Right now the Inventory Helper class (https://github.com/WayofTime/BloodMagic/blob/1.18.2/src/main/java/wayoftime/bloodmagic/util/helper/InventoryHelper.java) only has a single 'getAllInventories' method as when I wrote it that was the only check I needed a the time.  Some kind of 'getActiveSlots' method would need to be added, preferably with a way in the API for add-ons to register which of their inventory providers should be considered 'active' as well (ex. "curiosInventory" from `BloodMagicAPI.INSTANCE.registerInventoryProvider("curiosInventory", player -> getCuriosInventory(player));`).

stellanera98:
Added `registerActiveInventoryProvider(String)` to `IBloodMagicAPI` and `BloodMagicAPI` (since its just a list of strings maybe it shouldnt have provider in the name?)
Added `getActiveInventoryProvider()` to `BloodMagicAPI`
Added `getActiveInventories(Player)` to `InventoryHelper`
Changed `CuriosCompat.registerInventory()` to also register the inventory as active
Changed `ElementDivinedInformation.shouldRender()` to use `InventoryHelper.getActiveInventories(player)` instead of the code I copied from `CuriosCompat` before

src/main/java/wayoftime/bloodmagic/client/hud/element/ElementDivinedInformation.java
```diff
@@ -25,39 +26,20 @@ public ElementDivinedInformation(int lines, boolean simple, Class<T> tileClass)
...
-					flag = isFlagSigilHolding(sigilStack, true);
-			}
+		NonNullList<ItemStack> inventory = InventoryHelper.getActiveInventories(player);
 
-		} else
+		boolean flag = false;
+		for (int i = 0; i < inventory.size(); i++)
 		{
-			if (sigilStack.getItem() == BloodMagicItems.SEER_SIGIL.get())
+			ItemStack sigilStack = inventory.get(i);
```
VT-14: Optional: This For loop and `inventory.get(i)` could be replaced with an Enhanced For loop.  
`for (ItemStack sigilStack : inventory)`

I have also noticed that the Seer's Sigil in a Sigil of Holding does not work with the Incense Altar.  I've replicated the problem on the main branch so it isn't caused by this PR, but might be worth looking into.

VT-14: It looks like the Seers Sigil in a Sigil of Holding issue for the Incense Altar is fixed by removing the `&& !simple` check for the Seer's Sigil on line 56.

The Seer's Sigil only has a "simple" overlay, so that check is what is throwing it out when in the Sigil of Holding, where as it doesn't get thrown out when it is out of the Sigil of Holding.  The Seer's Sigil should be capable of showing both Simple and Non-Simple overlay info.

This does show another known issue where both the Simple and non-Simple overlays get rendered at the same time with the Seer's Sigil, but that is only a problem on the Blood Altar and only when the player has moved the overlay themselves (they default to being perfectly aligned on top of each other so is not a problem).  Changing that should likely be a separate PR.

VT-14 (approved):

src/main/java/wayoftime/bloodmagic/client/hud/element/ElementDivinedInformation.java
stellanera98: Adjusted the for loop and removed the mentioned `&& !simple` check

Is there a reason not to just remove the simple part from the non-simple overlay? Would probably also need to adjust the default position its rendered at

VT-14: That is how I would start trying to address that issue, but IMO that starts to exceed the scope of an expanded Curios Compat PR.  When I first started using GitHub I was told that it is best to try to keep each Pull Request focused on one purpose.

I only brought up the Sigil of Holding + Incense Altar issue because I was pretty certain the fix would be a small error in the file you were already editing, and it directly affected the Seer's Sigil usability as a Curio.

If you want to tackle the double-overlay issue then I would suggest making it a separate PR (it shouldn't affect curios compatibility).  You might want to move the `&& !simple` check to that one if you do make it.

I think this PR is good to go as it is now, though I'm leave it up to WayofTime (or TehNut [Nick]) for final approval/Pull.  Way has been pretty busy so I don't know how long that will take.

PR #1964, opened 2023-08-17
Fixing Green Growth ritual to plant seeds with Steadfast will

stellanera98:
- Added nether wart support on green growth ritual since it works on sugar cane and cactus too
- Changed `Utils.plantItemStack()` in order to work
It appears that IPlantable is no longer implemented on the items but the blocks instead. Saplings and flowers are also IPlantable 
- Removed the outmost two for loops and an if statement in above method because I think its functionally the same but I might be missing something


WayofTime:
Will reimplement this in 1.20 soon(tm)

PR #1976, opened 2023-09-27
Removed Some Log Spam

VT-14:
Changed several unused Models to use the new Resource Locations.  The previous ones had several UpperCase letters which are no longer valid and were throwing "Invalid Resource Location" errors.  The only models which didn't have assets in the new location were the Routing Nodes, which appear to have a different Model already.  Consider removing their unused models?

~~Made the Meteor Recipe check to see if a listed Mod ID is loaded.  This "modid" is optional and will use our ID (which is obviously loaded) if not listed.  If the supplied Mod ID is not loaded the recipe supplier returns null, which skips it with a single line in the log.  Added Mod IDs to the 5 mod-compat Meteors based on what was used for 1.16 documentation.~~

New: The Mod-Compat Meteor Recipes were changed to use Forge's Conditional loading system.  The Mod IDs used were pulled from the 1.16 documentation.

These changes remove 31 Stack Traces during startup (Invalid Resource Location), and 4 more on World Join and Data Pack Reloads (invalid recipe missing item; the AE2 one used a Tag that didn't throw an error) if those mods are missing (invalid recipe missing item), removing nearly 1,000 lines of log spam.

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/common/recipe/serializer/MeteorRecipeSerializer.java
+++ b/src/main/java/wayoftime/bloodmagic/common/recipe/serializer/MeteorRecipeSerializer.java
@@ -14,6 +14,8 @@ import net.minecraft.world.item.crafting.Ingredient;
 import net.minecraft.network.FriendlyByteBuf;
 import net.minecraft.util.GsonHelper;
 import net.minecraft.resources.ResourceLocation;
+import net.minecraftforge.fml.ModList;
+import wayoftime.bloodmagic.BloodMagic;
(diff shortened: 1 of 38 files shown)


dphaldes:
I don't think the modid checks will work, as the recipes will still be registered/loaded in. Instead, you should use [conditional loading](https://docs.minecraftforge.net/en/1.18.x/resources/server/conditional/#mod-loaded) as the recipes will not be loaded at all and will be completely handled by forge

VT-14:
The requested change has been made.

PR #1979, opened 2023-09-29
Update Spectral Block/Tile

### LINKED ISSUE #1978: Sigil Of Suppression breaks chunks
#### What happens:

Sigil of suppression lags a chunk and sometimes even chunk-bans players

Drain the water from a flooded base

#### Steps to reproduce:

1. Activate Sigil Of Suppression underneath water, in an area that is flooded with lots of water (An underwater base)
2. Turn off sigil/teleport away
3. Go back to original area where sigil was used, and there may be a bugged chunk (not sure how often it happens)
____

- BloodMagic: 3.3.1-43
- Minecraft: 1.20.1
- Forge: 47.1.3

VT-14:
`BlockSpectral.tick` was no longer being called.  Set up `BlockSpectra.getTicker` and moved tick logic to `TileSpectral.tick`.  This makes the Spectral Block decay back into the fluid block.

Currently needs review to ensure it works as intended.  Other changes may be required (ex. `canBeReplaced` is depreciated).

Fixes #1978.

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/common/block/BlockSpectral.java
+++ b/src/main/java/wayoftime/bloodmagic/common/block/BlockSpectral.java
@@ -3,12 +3,14 @@ package wayoftime.bloodmagic.common.block;
 import java.util.Random;
 
 import net.minecraft.core.BlockPos;
-import net.minecraft.server.level.ServerLevel;
 import net.minecraft.world.item.context.BlockPlaceContext;
 import net.minecraft.world.level.BlockGetter;
+import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.EntityBlock;
 import net.minecraft.world.level.block.entity.BlockEntity;
+import net.minecraft.world.level.block.entity.BlockEntityTicker;
+import net.minecraft.world.level.block.entity.BlockEntityType;
 import net.minecraft.world.level.block.state.BlockState;
 import net.minecraft.world.level.block.state.StateDefinition;
 import net.minecraft.world.level.block.state.properties.EnumProperty;
@@ -30,22 +32,14 @@ public class BlockSpectral extends Block implements EntityBlock
 		super(prop);
 	}
 
-	public void tick(BlockState state, ServerLevel world, BlockPos pos, Random rand)
-	{
-		switch (state.getValue(SPECTRAL_STATE))
-		{
-		case SOLID:
-			world.setBlock(pos, state.setValue(SPECTRAL_STATE, SpectralBlockType.LEAKING), 3);
-			world.scheduleTick(pos, this, BlockSpectral.DECAY_RATE);
-			break;
-		case LEAKING:
-			BlockEntity tile = world.getBlockEntity(pos);
+	@Override
+	public <T extends BlockEntity> BlockEntityTicker<T> getTicker(Level level, BlockState state, BlockEntityType<T> type) {
+		return (level1, blockPos, blockState, tile) -> {
 			if (tile instanceof TileSpectral)
 			{
-				((TileSpectral) tile).revertToFluid();
+				((TileSpectral) tile).tick();
 			}
-			break;
-		}
+		};
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


dphaldes:
Why is the functionality moved to the BE ? I don't think this requires a BE

VT-14:
It has been two weeks since I did it, but IIRC it is more consistent with the other Block Entities (Tiles) in the mod (I looked at the Hellfire Forge when trying to work out how it is supposed to work), and the BE had the data like State, Position, and Level already in it.

dphaldes:
oh okay! Then it makes sense. Move the blockEntity type check from block to BE for consistency as well (Other mods do it in BE as well)

WayofTime:
While this does technically work, the fluid seems to revert way too quickly. Ideally we would want a solution that makes it not revert at all while the player is around, such as by resetting the decaying process. This was done previously by having it handled by the TileEntity (are they called Block Entity or BE now?), and the sigil would refresh the TE's progress.

src/main/java/wayoftime/bloodmagic/common/tile/TileSpectral.java
```diff
@@ -82,4 +86,23 @@ public CompoundTag serialize(CompoundTag tag)
...
+
+	public void tick()
+	{
+		BlockState state = this.getBlockState();
+		BlockPos pos = this.getBlockPos();
+		Level level = this.getLevel();
+		switch (state.getValue(SPECTRAL_STATE)) {
+			case SOLID -> {
+				level.setBlock(pos, state.setValue(SPECTRAL_STATE, SpectralBlockType.LEAKING), 3);
+				level.scheduleTick(pos, state.getBlock(), BlockSpectral.DECAY_RATE);
```
dphaldes, line 98: `scheduleTick` is not required anymore

src/main/java/wayoftime/bloodmagic/common/tile/TileSpectral.java
```diff
@@ -82,4 +86,23 @@ public CompoundTag serialize(CompoundTag tag)
...
 	}
+
+	public void tick()
+	{
+		BlockState state = this.getBlockState();
+		BlockPos pos = this.getBlockPos();
+		Level level = this.getLevel();
+		switch (state.getValue(SPECTRAL_STATE)) {
+			case SOLID -> {
+				level.setBlock(pos, state.setValue(SPECTRAL_STATE, SpectralBlockType.LEAKING), 3);
```
dphaldes, line 97: Instead of '3' use the proper flag (Block.UPDATE_ALL). They are present in `Block.java`.

src/main/java/wayoftime/bloodmagic/common/block/BlockSpectral.java
```diff
@@ -30,22 +32,14 @@ public BlockSpectral(Properties prop)
...
-			break;
-		case LEAKING:
-			BlockEntity tile = world.getBlockEntity(pos);
+	@Override
+	public <T extends BlockEntity> BlockEntityTicker<T> getTicker(Level level, BlockState state, BlockEntityType<T> type) {
+		return (level1, blockPos, blockState, tile) -> {
 			if (tile instanceof TileSpectral)
 			{
-				((TileSpectral) tile).revertToFluid();
+				((TileSpectral) tile).tick();
```
dphaldes, line 40: you should pass the blockPos, level and tile from the lambda into the `tick` method instead of fetching them in the method. Also use the automatic cast when using `instanceof` instead of manual casting

dphaldes (changes requested):

src/main/java/wayoftime/bloodmagic/common/tile/TileSpectral.java
```diff
@@ -82,4 +86,23 @@ public CompoundTag serialize(CompoundTag tag)
 		tag.put("BlockState", NbtUtils.writeBlockState(storedBlock));
 		return tag;
 	}
+
+	public void tick()
+	{
+		BlockState state = this.getBlockState();
```
dphaldes, line 92: The reason the blocks vanish so quickly is that the BE is ticked every tick. Instead, use a tick_counter and check against the DECAY_RATE and only then tick it.

dphaldes (changes requested):

wrincewind:
closing this as stale - plus i'm reasonably sure it's fixed (in the latest batch of PRs, if not current). If i'm wrong, feel free to re-submit or re-open.

PR #1987, opened 2023-10-03
fix soul network reference for single player

### LINKED ISSUE #1982: Soul Network LP Transfers Between Single-Player Worlds
#### Issue Description:
Soul Network data appears to be transferring between Single Player worlds.  I have not tested Server interactions.

#### What happens:
Generate LP in one world, consume it in another.

Soul Network data to be save-specific (I would assume saved in the player.dat file).

#### Steps to reproduce:

1. Open World 1.  Use a Blood Orb.  Check Soul Network level with Divination Sigil.
2. Open World 2.  Use Divination Sigil and see that you have the same level of LP without ever generating it in that world.
3. Use another Sigil (ex. Lava Sigil) to confirm that you are able to use that LP in that world.
4. Return to World 1 and see that the LP level has decreased based on the usage in World 2.

____

- BloodMagic: 3.3.1-43.
- Minecraft: 1.20.1
- Forge: 47.1.3

dphaldes:
Fixes #1982 
adds a variable that caches the reference to DimensionDataStorage and updates savedDatahandler when references are updated

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/util/helper/NetworkHelper.java
+++ b/src/main/java/wayoftime/bloodmagic/util/helper/NetworkHelper.java
@@ -9,7 +9,11 @@ import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.level.storage.DimensionDataStorage;
 import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.event.server.ServerStoppedEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
 import net.minecraftforge.server.ServerLifecycleHooks;
+import wayoftime.bloodmagic.BloodMagic;
 import wayoftime.bloodmagic.common.item.BloodOrb;
 import wayoftime.bloodmagic.common.item.IBindable;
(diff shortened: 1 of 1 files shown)


src/main/java/wayoftime/bloodmagic/util/helper/NetworkHelper.java
```diff
@@ -33,16 +36,14 @@ public class NetworkHelper
...
-		{
-			if (ServerLifecycleHooks.getCurrentServer() == null)
-				return null;
+		if (ServerLifecycleHooks.getCurrentServer() == null)
+			return null;
 
-			DimensionDataStorage savedData = ServerLifecycleHooks.getCurrentServer().overworld().getDataStorage();
+		DimensionDataStorage savedData = ServerLifecycleHooks.getCurrentServer().overworld().getDataStorage();
 
+		if (dataHandler == null || cachedDataStorage != savedData)
```
VT-14: `cachedDataStorage` is never assigned a value, so this is effectively `if (dataHandler == null || savedData != null)`.

dphaldes: bruh how did I miss that. I will fix that. 
How did it work during my testing then lol

PR #1993, opened 2023-10-11
Added Error Flag to Hellfire Forge

VT-14:
Added an error flag to the Hellfire Forge that tells players when the recipe isn't working due to not having enough Demon Will on hand.

* Added flag texture to the Hellfire Forge's GUI.  Also saved the texture more efficiently, decreasing file size.
* Added translation keys for tooltip text.
* added a Boolean variable to the Tile, and a method to read it.
* added the flag and tooltip to the Screen.

I also tweaked the TileSoulForge's operation.  The `|| burnTime > 0` check caused a minor bug where a player could start the recipe going, then swap to a gem without enough will and it would look like it would keep going (though fail to craft when finished).  It seems to work as I would expect without it.

I then could also remove two redundant if statements checking for the level being not client side, and having enough Will.  the client side check was done earlier so was always true, and by removing the `burnTime > 0` alternative the minimum souls check is also already done and always true.


wrincewind (approved):
tested and works as expected. thanks again vt!

PR #1995, opened 2023-10-13
Advancements

VT-14:
Adding Advancements to Blood Magic.


wrincewind:
while i'm going through all of the optional updates and bugfixes and such - any thoughts on this? Do you want to add some more achievements, have us add it to 1.20 as-is, or hold off for the eventual 1.21 port?

wrincewind:
I definitely do still like the idea of having advancements in BM, but i'm considering closing this as stale - we can re-visit it for 1.21 (unless you (or anyone else that wants to chip in) think you can whip together a reasonably feature-complete version in a short amount of time :p)

PR #2000, opened 2023-10-24
[1.20.1] Makes Living Evolution configurable

dphaldes:
`LivingStats` tag now contain a boolean `evolved` that stores whether the Item has evolved or not. Living evolution ritual now simply calls the `setEvolved` method in `LivingStats` which updates the tag to true. By default, this also updates the `maxPoints` to 300, but this can be overridden for custom behaviour.
For example, One could remove the points update during evolution for custom item but still retain the information that item has evolved or not.
`ILivingContainer` also now contains a method to control whether evolution is allowed or not.


wrincewind (approved):
going on discord discussions - it'd be good if this were definable in the config files - as far as i can tell this only ("only") opens the values up for tweaking by other mods? But, even if so, that'd be out-of-scope for this update.

PR #2005, opened 2023-11-09
fix spectral sight with mixins

dphaldes:
Reimplements the Spectral sight mixins. 
Mixins work now and also added MixinExtras (it's more pleasant to use)


wrincewind (approved):
can't say much more than 'tested it, works as expected, great job'

PR #2008, opened 2023-11-09
fix checks for altar structure

### LINKED ISSUE #1999: [1.20.1 Forge] Chisel and bits blocks not working for Tier III and IV altar
#### Issue Description:
Just built a Tier III altar and I was trying to decorate the pillars with chisel-and-bits netherbrick with lava, but then, when I was going to use the altar, it was still in Tier II.

The pillars:
image
The altar:
image

#### What happens:
The altar does not go to Tier III if I use any block edited with chisel-and-bits

The altar go to Tier III, as in older versions this feature worked flawlessly 

#### Steps to reproduce:

1. Build a Tier III altar
2. Use any block affected with chisel-and-bits as pillars for the glowstone
3. Check the altar with the Divination Sigil

____

- BloodMagic: 3.3.2-44
- Minecraft: 1.20.1
- NeoForge: 47.1.53

dphaldes:
fixes #1999

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/altar/AltarUtil.java
+++ b/src/main/java/wayoftime/bloodmagic/altar/AltarUtil.java
@@ -33,7 +33,7 @@ public class AltarUtil
 				BlockPos componentPos = pos.offset(component.getOffset());
 				BlockState worldState = world.getBlockState(componentPos);
 
-				if (component.getComponent() == ComponentType.NOTAIR && worldState.isSolid())
+				if (component.getComponent() == ComponentType.NOTAIR && !worldState.isAir() && !worldState.liquid())
 					continue;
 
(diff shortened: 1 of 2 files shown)


wrincewind (changes requested):
right now, this seems a little too permissive, where the previous was a little too restrictive - items without collision (such as vines, torches, signs, sugarcane) are considered valid blocks for this. I think signs are the bare minimum that makes sense here. Of course, we want to encourage creative builds, and i don't know how these restrictions would affect e.g. Chisels And Bits blocks, but is there some middle-ground we can work towards?

wrincewind:
after some discussion on Discord about the feasability of such, it's probably better to err on the side of 'too permissive' rather than try and get it exactly right and risk causing issues with other decorative mods / limit player choice. If you want to balance glowshroom blocks on top of bamboo for your altar, go nuts.

wrincewind (approved):

PR #2046, opened 2024-06-21
fix config issues with self sacrifice options

stellanera98:
removed old config remnant
updated references to use the new config

fixes an issue where ritual of feathered knife and sacrifices while "incensed" would always use 100 as sacrificialDaggerConversion, even if a different amount was set in the config


wrincewind (changes requested):
I think we should keep in ConfigHandler for now as effectively one big ToDo. If nothing else, for reference in later versions. Unless you think you can properly re-implement it and replace it with working code :p

stellanera98:
could you open an issue for the config things we figured could be made for 1.20, tag/assign me on it and merge this one? or whats your opinion on this now?

wrincewind:
https://github.com/WayofTime/BloodMagic/issues/2100 here you go!

PR #2047, opened 2024-06-24
fix filters saving on wrong itemStack

### LINKED ISSUE #2041: Standard filter stacking
#### Issue Description:
Found in Stoneopolis modpack, tested in an instance with just Blood Magic, Patchouli and JEI and still happens
There is a problem with stacking standard filters. When you unstack two filters, they are still linked in a buggy way.
I'm guessing this bug has to do something with pointers.

#### What happens:
When two filters are unstacked, only the one in the original slot gets affected.

The two filters should behave seperately

#### Steps to reproduce:

1. get a stack of 2 standard filters in the hotbar
2. set items in both filters while they are stacked in one slot and close the GUI
3. split the stack into another hotbar slot using right click in the inventory GUI
4. open the second filter GUI and change something
5. close the GUI and watch the first filter being affected and the second one not

____

- BloodMagic: 3.3.3-45
- Minecraft: 1.20.1
- Forge: 47.2.0

stellanera98:
fixes #2041

Filter items no longer set a UUID on use
The saving mechanism no longer checks for the UUID and uses slot index instead


stellanera98:
Added recipes for all filter kinds to clear their settings.
Tested by pulling out two fresh filters of a kind from the creative menu, setting both of them to something different which makes them unstackable. Place both of these by themselves into a crafting grid which clears their stored data and makes them stack with each other and new filters again

wrincewind (approved):
gonna have to put an entry for this in the Sanguine Scientum, but otherwise, great!

Issue #2034, opened 2024-04-01
Reinforced variants of runes have no effective tool


cEMRSS:
#### Issue Description:
All reinforced variants of runes (e.g. "Reinforced charging rune") have no effective tool, and the drop requirements either do not exist or cannot be met.

#### What happens:
This means that when these blocks are broken in survival, they never drop their corresponding item, regardless of the tool used.

#### What you expected to happen:
One would expect to be able to break a block that has a dedicated un-crafting recipe, and get the item back.

#### Steps to reproduce:
Attempt to break a reinforced charging rune while in survival with a netherite pickaxe.

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: 1.20.1-3.3.2-44
- Minecraft: 1.20.1
- Forge: 47.2.16

cEMRSS:
I will note that using the setblock or fill commands with the "destroy" keyword does drop the relevant item.

Issue #2023, opened 2024-02-05
Life Essence Fluid doesn't use localized name in JEI


MrNoe93:
#### Issue Description:
Life Essence Fluid in JEI does not use its localized name.

#### What happens:
When searching for life essence in JEI, the fluid shows up as "life_essence_fluid_type"

#### What you expected to happen:
Life Essence Fluid in JEI to be named Life Essence Fluid

#### Steps to reproduce:
Install JEI and Blood Magic and search for life essence in JEI.  Hover over the fluid.

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: 3.3.2-44
- Minecraft: 1.20.1
- Forge: 47.2.20

PR #2048, opened 2024-06-24
fluid names & reinforced runes tools

stellanera98:
- corrected fluid translation keys so life essence and doubt fluids have an actual name #2023
- added tool and tier needed tags for reinforced runes so they now actually drop themselves when broken with correct tools #2034
- updated the altar recipe in the generator to reflect changes to the generated json from wrince in 983fc0666b3e6607c19fa84041565d33dd1cdd42
- ran datagen to make the changes effective

I figured stone is a bit low for something made with netherite so I made the reinforced runes need dia level pickaxes


wrincewind (approved):
all seems to work OK. the Bucket of Doubt is still referred to by its unlocalised name - I'll see if I can't do something about that.

stellanera98:
for consistencies sake Ill put it here as well:
Bucket of Doubt's translation would be something like `addItem(BloodMagicItems.DOUBT_BUCKET, "Bucket of Doubt")` inside the wayoftime/bloodmagic/common/data/GeneratorLanguage.java file's addTranslations method and then running the data generation task

PR #2049, opened 2024-06-24
stop teleposer from crashing when activated with unbound focus in it

### LINKED ISSUE #2028: Teleposers are crashing my game
#### Issue Description:
The Teleposer crashes the server when I use it with one unbound teleposition focus.

#### What happens:

The server crashes

I teleport to the other teleposer

#### Steps to reproduce:

1.  Place one teleposer and put a focus in it.
2. Bind a focus to the teleposer.
3. Place another teleposer elsewhere and put the second focus in it. 
4. Stand on top of the second teleposer and power it
...

____

- BloodMagic: 3.3.2-44
- Minecraft: 1.20.1
- NeoForge: 47.1.99

stellanera98:
fixes #2028 though I only managed to get a crash when the source has an unbound focus, not the destination. It doesnt seem like the destination is used at all except to make sure there is a teleposer there
- add a check for the binding not to be null before using it

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/common/tile/TileTeleposer.java
+++ b/src/main/java/wayoftime/bloodmagic/common/tile/TileTeleposer.java
@@ -196,9 +196,12 @@ public class TileTeleposer extends TileInventory implements MenuProvider, Comman
 	private SoulNetwork getNetwork()
 	{
 		ItemStack focusStack = this.getItem(FOCUS_SLOT);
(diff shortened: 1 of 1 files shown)


wrincewind (approved):
Looks to solve the problem! No changes needed. It would be more user-friendly if the teleposer checked for whether the focus was bound or not on insertion and block it if un-bound, but that's beyond the scope of this PR.

PR #2069, opened 2024-11-08
Fix ARC not respecting unbreakable items

### LINKED ISSUE #1754: Do not Try to Damage Unbreakable Items in ARC
#### Issue Description:
Unbreakable items (tested with Explosive Powder) put into the ARC Disapear or take damage after processing
#### What happens:
Tested Origionaly with Forbidden Arcanus's Indestructable enchantment the Explosive power continued to take damage despite it not suppost to. I then spawned in a Explosive Powder with the Vanilla unbreakable tag with /give @p bloodmagic:explosivepowder{Unbreakable:1} 1 and as soon as the item processed 1 item it disappeared

Items to Not take damage and not to disapear

#### Steps to reproduce:

1. Enchant Explosive Powder with a Stella from forbidden magic to make it unbreakable Or give yourself one with the unbreakable tag
2. Put it in the arc furnace (in this case with netherrack)
3. Watch it take durrability damage or Disapear depending on which is used.
...

____

- BloodMagic: 3.0.5-10
- Minecraft: 1.16.3
- Forge: 35.1.36

stellanera98:
ItemStack.isDamageableItem returns false if the NBT tag Unbreakable is set to true so it should not be used to check if the item theoretically has durability, Item.isDamageable(ItemStack) returns true if the item has maxDamage set to anything more than 0 so it should return true for anything that would have durability, even if Unbreakable is set to true (or a Forbidden & Arcanus Eternal Stella is applied)
Fixes #1754

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/common/tile/TileAlchemicalReactionChamber.java
+++ b/src/main/java/wayoftime/bloodmagic/common/tile/TileAlchemicalReactionChamber.java
@@ -381,10 +381,10 @@ public class TileAlchemicalReactionChamber extends TileInventory implements Menu
 		ItemStack toolStack = getItem(ARC_TOOL_SLOT);
 		if (!toolStack.isEmpty())
 		{
-			if (toolStack.isDamageableItem())
+			if (toolStack.getItem().isDamageable(toolStack))
 			{
 				int unbreakingLevel = EnchantmentHelper.getItemEnchantmentLevel(Enchantments.UNBREAKING, toolStack);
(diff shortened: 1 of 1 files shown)


wrincewind:
I've tested this with the vanilla unbreaking tag. and, though it took a while, also confirmed it works with Forbidden & Arcanus' Eternal Stella. It'd be nice if it being unbreakable could remove the 'uses remaining' line from the tooltip, but i think that can wait for a different PR.

wrincewind (approved):

PR #2070, opened 2024-11-10
Create zh_cn.js

Dumpl1n9:
a lang for chinese players


dphaldes:
This needs to be a json file not a js file

PR #2077, opened 2025-02-22
change imagen size 

SrNadien:


wrincewind:
Before we merge this - what is the purpose of this change? Thaumcraft isn't available in 1.20.1 and this texture wasn't used for anything else, as far as we know. (Also, why is it a .html file now?)

dphaldes:
This PR looks very sketchy. The added file isn't even a plain text html. Since the image is not even used anymore, it might be a good idea to close this PR @wrincewind

wrincewind:
Agreed - i thought as much but wanted to give them time to explain their reasoning. (and wondered if maybe the html was a weird accident)

SrNadien:
@wrincewind delete all textures thaumcraft

PR #2089, opened 2025-07-21
Harvest Ritual fix 1.20

stellanera98:
Replace the reflection calls with actual API calls.
To test, uncomment the MAG and Cucumber dependencies in build.gradle to get MAG running in Dev, place harvest ritual, place some MAG crops and watch as it harvests them once they are fully grown

If this ritual should be working with other harvestables like kelp or even things like amethysts those would be addable as well


wrincewind:
I'm not sure about kelp... does the current ritual harvest sugarcane? if so it should do kelp as well, but given both can be harvested with pistons, i'm not *so* sold on it. 

As far as harvesting amethyst and other 'budding' blocks, if anything I feel those should be harvested by the "Crack of the Fractured Crystal" ritual. Or maybe a new ritual specfically for budding blocks? in either case, we'd need to update the info in the language files to reflect that - generatorlanguage.java#L334.

stellanera98:
current version harvestables: Carrots, Wheat, Potatoes, Beetroot, Pumpkin, Melon, Sugarcane, Cactus, Sweet Berries, Glow Berries, Nether Wart, both Tau variants as well as the semi implemented Growing Doubt. Along with a few crops from third party mods like pams harvestcraft (didnt test those) and some others Im not even sure are in 1.20

wrincewind (changes requested):
it doesn't quite work how i'd expect with glowberries - as per the wiki, the berries can only grow when a new length of vine grows, so i'd expect the vine to be broken when it grows, regardless of whether it has a berry on it or not. (similar to kelp/sugarcane/bamboo)

I can see an edge case where a player might want to use the green grove ritual to bonemeal the vines to force-grow glow berries, but the Green Grove ritual doesn't currently seem to work on them - possibly something to look into? We could always add a Steadfast Will Aspect to the ritual to make it not break the vines, but that does seem quite niche. 

in short, it should definitely break glow-vines when they grow regardless of berry presence, beyond the first vine, it should maybe have a will aspect if that's not too hard to add in, and we should probably double-check the green grove ritual.

wrincewind:
i didn't have time to check the other weirder stuff like dripleaves and the other cave plants - making a note for myself tonight/tomorrow. it's all looking good otherwise though!

wrincewind (approved):

PR #2090, opened 2025-07-21
Add filter copy recipe

stellanera98:
Put filter to copy from in the first slot (top left) of any crafting grid with at least two slots total. Place any of the same kind of filter in any of the other slots. Result will be the amount of total filters in the grid with the same data on it as the "primary" (first slot) had


wrincewind (approved):
time spent testing: 5 minutes. Time spent documenting: 4 hours.

PR #2091, opened 2025-07-21
fix HUD element saving

### LINKED ISSUE #2050: HUD location adjustments not saved by default
#### Issue Description:
Shift + Right Click with an item that provides a HUD opens a GUI to move them around so you have the option to put them whereever you want them to be. This works great until you leave the world. The location this is supposed to save to is config/bloodmagic/hud_elements.json but the config/bloodmagic folder does not exist anymore and so the path cannot be found. Creating the bloodmagic folder manually solves this.

I guess the simplest way to fix that would be to have it save to config/bloodmagic-hud_elements.json instead. The config folder already exists (or is created by mc/forge) and its more in style with current config naming. Or have forge/mc create the bloodmagic folder, however currently hud_elements.json would be the only file in there and I dont know if the -client and -server.toml can be moved. The forge doc doesnt suggest so.

#### What happens:
The HUD positions configured reset to default for each individual session since it cannot save it

The HUD positions to be persistent even between sessions in the same world (or instance rather)

#### Steps to reproduce:

1. Open the HUD config screen by shift+right click with any of the items (holding/divination/seer sigil, demon will aura gauge)
2. Rearrange the elements in some way
3. Click the save button
4. Notice there is an error printed in the console
5. Leave the world
6. Open the world again
7. Notice the elements are in their default position
8. Create the bloodmagic folder in the config folder
9. Repeat 1-6, no error in console and elements stay in the rearranged positions

___

- BloodMagic: 3.3.3-45
- Minecraft: 1.20.1
- Forge: 47.1.0

### LINKED ISSUE #2084: Crash with customised HUD for Seer's Sigil and Divination Sigil
partly listing this as a known workaround so people can solve this issue if they come across it, partly reporting a bug. 

In some versions of BM, if you attempt to customise your UI with the Seer's Sigil or Divination Sigil (shift-right-click with the sigil in hand, drag around the various boxes, hit save), the game will crash.

This is due to BM attempting to save a file to the folder /config/bloodmagic/ on the client-side, and if this folder is not present, rather than creating it, it crashes. The workaround is to create an empty 'bloodmagic' folder in the config folder, then re-attempt saving your layout.

versions affected: 
Bloodmagic-1.20.1-3.3.3-45.jar (maybe others)

stellanera98:
fixes #2050 fixes #2084 
Makes sure the config/bloodmagic folder exists so the config for the HUD element positions can be saved there

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/client/hud/ElementRegistry.java
+++ b/src/main/java/wayoftime/bloodmagic/client/hud/ElementRegistry.java
@@ -96,6 +96,7 @@ public class ElementRegistry
 			toWrite.put(entry.getKey().toString(), entry.getValue().getPosition());
 
 		String json = GSON.toJson(toWrite);
+		FMLPaths.getOrCreateGameRelativePath(FMLPaths.CONFIGDIR.get().resolve(BloodMagic.MODID));
 		try (FileWriter writer = new FileWriter(CONFIG))
 		{
 			writer.write(json);


wrincewind (approved):
works just fine, it looks like. Cheers!

PR #2092, opened 2025-07-21
Fix filters having duplicate NBT

stellanera98:
Seems to work fine if the old format is encountered.
Also touched up the filter copy entry


wrincewind (approved):
looks to work fine - since this touched copying_filters.json i took the liberty of updating that manual entry at the same time (it was on my to do list).

PR #2095, opened 2025-07-25
fix: save te data

UADominator:
In some cases, the inventory doesn't have time to save, which leads to dupe.


wrincewind:
Can you be more specific? (Mostly to help us with testing this and ensuring that it 1) works as expected and 2) doesn't break anything else?) 

- which inventory?
- what cases?
- what's being duplicated? 

if you could give us a step-by-step of how to reproduce the bug, we can try it with and without the code-change to ensure that the fix worked. 

Thanks in advance!

UADominator:
Which inventory: Blood Altar
In which cases: The item from the Blood Altar is taken, after which the chunk quickly unloads (portal teleport / /home command, etc.)
Not on all machines does the duplication occur, because there is enough time to save the TE data to the world file before the chunk unloads

dphaldes (approved):

dphaldes:
This is required. It is a common bug I encountered when building ToB

PR #2108, opened 2025-10-24
fix comparator interaction

### LINKED ISSUE #2102: Crash on comparator placement.
The game crashes when a comparator is placed next to an altar when a bloodstone block is underneath it.  The problem has been speculated to be this line:

"return NetworkHelper.getSoulNetwork(binding).getCurrentEssence() * 15 / orb.getCapacity();"

Being that it may be possible that "orb.getCapacity()" is returning 0, and thus, the return is dividing by zero.

Something like this should be implemented instead:

float capacity = orb.getCapacity();
if (capacity <= 0) {
    return 0;
}
return NetworkHelper.getSoulNetwork(binding).getCurrentEssence() * 15 / capacity;

I will note that the crash happens before the comparator is placed, and the crash actually prevents the comparator from being placed in the world.  The crash can be mitigated by placing the comparator first, then placing the bloodstone afterwards.

stellanera98:
fixes #2102

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/altar/BloodAltar.java
+++ b/src/main/java/wayoftime/bloodmagic/altar/BloodAltar.java
@@ -7,6 +7,7 @@ import net.minecraft.core.particles.DustParticleOptions;
 import net.minecraft.core.particles.ParticleTypes;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.server.level.ServerLevel;
+import net.minecraft.util.Mth;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.RedstoneLampBlock;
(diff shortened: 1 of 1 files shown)


src/main/java/wayoftime/bloodmagic/altar/BloodAltar.java
```diff
@@ -793,7 +796,12 @@ public int getAnalogSignalStrength(int redstoneMode)
 			if (binding == null || orb == null)
 				return 0;
 
-			return NetworkHelper.getSoulNetwork(binding).getCurrentEssence() * 15 / orb.getCapacity();
+            SoulNetwork network = NetworkHelper.getSoulNetwork(binding);
+            if (network == null)
+                return 0;
+
+            // see AbstractContainerMenu#getRedstoneSignalFromContainer
+            return Mth.floor((float) network.getCurrentEssence() / (float) orb.getCapacity() * 14.0F) + (network.getCurrentEssence() > 0 ? 1 : 0);
```
wrincewind: this needs to be clamped to cap at a maximum output of 15 - if you put in an orb with a value lower than the soul network's fill level, you can end up with the comparator producing redstone signals >15.

wrincewind (changes requested):

wrincewind (approved):

PR #2114, opened 2025-11-04
Hydration recipes

stellanera98:
tacking on the armour tags inclusion as well as adding the 2 missing tiers to the sentient tools (and moving the scythe to the others) since doing these separately seemed like overkill


wrincewind (changes requested):
immediate issue with all the concrete recipes: they turn concrete powder into... concrete powder. D:

wrincewind (approved):
i slightly messed up the name of a commit but i'm happy with this as it is now!

PR #2115, opened 2025-11-04
agricraft integration!

stellanera98:
you can uncomment lines 116 and 117 in build.gradle to enable the latest version of agricraft and its dependency in the runs for testing. you also might want to disable weeds in the agricraft config

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/ritual/types/RitualGeode.java
+++ b/src/main/java/wayoftime/bloodmagic/ritual/types/RitualGeode.java
@@ -72,7 +72,7 @@ public class RitualGeode extends Ritual {
         addBlockRange(CHEST_RANGE, new AreaDescriptor.Rectangle(new BlockPos(0, 1, 0), 1));
         addBlockRange(HARM_RANGE, new AreaDescriptor.Rectangle(new BlockPos(-1, -3, -1), 3, 3, 3));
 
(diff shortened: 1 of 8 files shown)


src/main/java/wayoftime/bloodmagic/ritual/types/RitualGeode.java
```diff
@@ -72,7 +72,7 @@ public RitualGeode() {
         addBlockRange(CHEST_RANGE, new AreaDescriptor.Rectangle(new BlockPos(0, 1, 0), 1));
         addBlockRange(HARM_RANGE, new AreaDescriptor.Rectangle(new BlockPos(-1, -3, -1), 3, 3, 3));
 
-        setMaximumVolumeAndDistanceOfRange(HARVEST_RANGE, 3 * 3 * MAX_BLOCKS, 15, 15);
+        setMaximumVolumeAndDistanceOfRange(HARVEST_RANGE, 2 + 2 * 3 * MAX_BLOCKS, 15, 15);
```
wrincewind, line 75: I gotta know... why?

wrincewind (approved):

src/main/java/wayoftime/bloodmagic/ritual/types/RitualGeode.java
stellanera98: Because I goofed doing math it seems. The idea is that if you have a line of budding blocks its always 2 more "slices" of 1x3x3 than there is blocks (the mentioned fence posts). So it should probably actually be 3 * 3 * (MAX_BLOCKS + 2)

PR #2136, opened 2025-12-11
Adds block protection for all BM world interaction

### LINKED ISSUE #2024: FTB Teams claim bypass with item from mod
#### Issue Description:
Heyo. While this is an issue that comes up in modpacks, I though I would let you guys know so that the cross mod issue might be fixed faster.

#### What happens:
Charges from BM do not respect claims from the FTB Teams mod

For charges not to be throwable into claims

#### Steps to reproduce:

1. Look at a block that is inside the claim while being far enough away to not interact with it
2. Your location inside or outside the claim does not matter
3. Right click while holding any above item
4. This causes a mod feature of being able to throw the item causing it to fly to the block you are looking at
5. Claim protections do not block placement of item
6. The item (now a block placed in the world) functions as if not in a claim at all.

____

- BloodMagic: 3.3.2-44
- Minecraft: 1.20.1
- Forge: 47.2.20
___
#### Post at the FTB Issues github
https://github.com/FTBTeam/FTB-Mods-Issues/issues/1094#issue-2128707576

Saereth:
Fixes #2024 Implemented block protection to prevent unauthorized block breaking and placing in protected chunks.

Introduces a BlockProtectionHelper class with methods to fire BlockEvent.BreakEvent and BlockEvent.EntityPlaceEvent before block modifications.

This allows protection mods like FTB Chunks or Xaeros Parties to prevent unauthorized operations by rituals, sigils, projectiles and other automated systems.

Adds owner tracking to explosive charges to allow protection against griefing.

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/ritual/types/RitualCrystalSplit.java
+++ b/src/main/java/wayoftime/bloodmagic/ritual/types/RitualCrystalSplit.java
@@ -12,6 +12,9 @@ import wayoftime.bloodmagic.api.compat.EnumDemonWillType;
 import wayoftime.bloodmagic.common.block.BloodMagicBlocks;
 import wayoftime.bloodmagic.common.tile.TileDemonCrystal;
 import wayoftime.bloodmagic.ritual.*;
+import wayoftime.bloodmagic.util.helper.BlockProtectionHelper;
+
+import java.util.UUID;
 
 import java.util.function.Consumer;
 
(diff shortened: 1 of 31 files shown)


src/main/java/wayoftime/bloodmagic/ritual/types/RitualCrystalSplit.java
```diff
@@ -110,10 +113,11 @@ public void performRitual(IMasterRitualStone masterRitualStone)
 
 			rawTile.setCrystalCount(rawTile.getCrystalCount() - 4);
 
-			growCrystal(world, vengefulPos, EnumDemonWillType.VENGEFUL, vengefulCrystals);
-			growCrystal(world, corrosivePos, EnumDemonWillType.CORROSIVE, corrosiveCrystals);
-			growCrystal(world, steadfastPos, EnumDemonWillType.STEADFAST, steadfastCrystals);
-			growCrystal(world, destructivePos, EnumDemonWillType.DESTRUCTIVE, destructiveCrystals);
+			UUID ownerUUID = masterRitualStone.getOwner();
```
dphaldes, line 116: Do we want to protect crystal growth? not exactly griefing anyone

dphaldes (approved):
Looks good! Dont see any major problems! 
Yet to test in game but should work fine!

src/main/java/wayoftime/bloodmagic/ritual/types/RitualCrystalSplit.java
Saereth: Yeah I was thinking of some weird case where you ritual tinker over and make someones crystals grow faster to mess with their will setup but I guess its enough of a niche thing its probably a non issue. Those crystals are going to grow regardless

stellanera98: This ritual doesnt have ranges to reconfigure.
https://github.com/WayofTime/BloodMagic/blob/1.20.1/src/main/java/wayoftime/bloodmagic/ritual/types/RitualCrystalSplit.java#L21-L24

also the way to stop that from happening should be to check if the player can build in that chunk and if they cant stop the tinkerer from working

Saereth: yeah might be nice to have something special for the tinkerer itself at some point

wrincewind:
Game launches and didn't run into any issues when testing. I'm not sure how exactly to go about testing in singleplayer - we may have to keep an eye out for reports from people running multiplayer servers about permissions issues.

Issue #2130, opened 2025-12-04
plantleach effect causes ticking entity crash when growing a tree


305pence:
context: if you give the green grove ritual corrosive will, it will give this effect to nearby mobs which causes plants near them to grow

how to recreate: 
1. spawn a pig
2. place a sapling nearby
3. /effect give @e[type=minecraft:pig] bloodmagic:plantleech
the game will crash (unless you have that one mod that tries to prevent ticking entity crashes)
note when you log in, the tree will be grown and the game will not crash until the above is repeated

log:
[message (3).txt](https://github.com/user-attachments/files/23924470/message.3.txt)
from this i assume it grows the tree, but then tries to "grow" the log where the sapling used to be which lacks the blockstates for that

wrincewind:
fixed by #2137 - will close with next release.

PR #2137, opened 2025-12-11
Fixed plant leech effect issue

Saereth:
#2130
The code now:
  1. Gets the BlockState fresh at each iteration of the inner loop
  2. Checks if the block has changed from the original block type
  3. Checks if the block is still a BonemealableBlock
  4. If either condition fails (block changed or no longer growable), it breaks out of the loop

  This prevents the crash when a sapling transforms into a tree mid-growth, as well as any other block transformations that might occur during random ticks.

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/potion/BMPotionUtils.java
+++ b/src/main/java/wayoftime/bloodmagic/potion/BMPotionUtils.java
@@ -51,19 +51,24 @@ public class BMPotionUtils
 
 		for (BlockPos blockPos : growList)
 		{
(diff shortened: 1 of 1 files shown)


wrincewind:
Looks solid to me - would you mind changing the description to not say 'Fixes #2130' but instead just '#2130'? I'll add a tag to the issue saying 'fixed in dev' and close it when we do the release. (Also if you could make this change to your other PRs, that'd be appreciated). 

the release'll probably be quite soon but this is something we should've been doing a while back, it keeps things straightened out and prevents duplicate issues being raised, confusion about what was fixed when, etc.

PR #2138, opened 2025-12-11
Conditionally registers Agricraft harvest handler.

Saereth:
Registers the Agricraft harvest handler only if the Agricraft mod is loaded.

Prevents a crash when Agricraft is not present.

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/ritual/ModRituals.java
+++ b/src/main/java/wayoftime/bloodmagic/ritual/ModRituals.java
@@ -1,6 +1,7 @@
 package wayoftime.bloodmagic.ritual;
 
 import net.minecraft.world.level.block.Blocks;
+import net.minecraftforge.fml.ModList;
 import wayoftime.bloodmagic.ritual.harvest.*;
 
 public class ModRituals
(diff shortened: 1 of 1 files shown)


wrincewind:
This does fix the problem in worlds where the crash hasn't yet occured, but when i to fix a world that was already bricked by the issue by updating to this version, it continues to crash with the same error (noClassDefFoundError: com/agricraft/agricraft/common/block/entity/CropBlockEntity'). 

Do you think you can include a fix to un-brick already broken worlds? if not we can update the changelog to include a note about how to solve this, like 'use the config setting removeErroringBlockEntities   in serverconfigs/forge-server.toml', but i'd prefer a more elegant solution if possible.

Saereth:
I can include a datafixer to sort this yeah, I'll get it in tomorrow since its getting late here today.

wrincewind:
as per discussion on discord, it'll be safer to keep this commit as-is and include this note in the changelogs: 
"in your world save directory, go to /serverconfig/forge-server.toml' and change 'removeErroringBlockEntities' to 'true', launch your world once, then change it back to false."

Issue #2085, opened 2025-06-01
Log Spam even without death now.


border999:
https://gist.github.com/border999/f73ae4189e65335fe71babf6415e749e -excerpt from log since gist.github doesn't like 70MBs of spammed logs.
bloodmagic-1.20.1-3.3.3-45
curios-forge-5.14.1+1.20.1
All other possibly relavent curio addons
convenientcurioscontainer-3.1-BETA-forge-1.20.1
mekanismcurios-1.20.1-1.2.0
ApothicCurios-1.20.1-1.0.3d
curious_armor_stands-1.20-5.1.0
curiotiab-1.20.1-2.0.2

border999:
If you have any questions or need more information, I will provide all I can within reason.

wrincewind:
fixed by #2139 and #2045 - will close on next release.

Issue #2040, opened 2024-05-25
i dont know how to do PR's lol Please update the CuriosCompat.java class to fix a crash with Corail Tombstone. (1.20.1)


Orangejewce92:
package wayoftime.bloodmagic.compat;

import net.minecraft.world.entity.player.Player;
import net.minecraft.world.item.ItemStack;
import net.minecraft.core.NonNullList;
import net.minecraftforge.fml.InterModComms;
import net.minecraftforge.fml.event.lifecycle.InterModEnqueueEvent;
import net.minecraftforge.items.IItemHandler;
import top.theillusivec4.curios.api.CuriosApi;
import top.theillusivec4.curios.api.SlotTypeMessage;
import top.theillusivec4.curios.api.SlotTypePreset;
import top.theillusivec4.curios.api.type.util.ISlotHelper;
import wayoftime.bloodmagic.BloodMagic;
import wayoftime.bloodmagic.core.LivingArmorRegistrar;
import wayoftime.bloodmagic.core.living.LivingStats;
import wayoftime.bloodmagic.core.living.LivingUtil;
import wayoftime.bloodmagic.impl.BloodMagicAPI;

import java.util.Optional;

public class CuriosCompat
{
    public void setupSlots(InterModEnqueueEvent evt)
    {
        InterModComms.sendTo("curios", SlotTypeMessage.REGISTER_TYPE, () -> SlotTypePreset.NECKLACE.getMessageBuilder().build());
        InterModComms.sendTo("curios", SlotTypeMessage.REGISTER_TYPE, () -> SlotTypePreset.CHARM.getMessageBuilder().build());
        InterModComms.sendTo("curios", SlotTypeMessage.REGISTER_TYPE, () -> new SlotTypeMessage.Builder("living_armour_socket").size(0).icon(BloodMagic.rl("item/curios_empty_living_armour_socket")).build());
    }

    public void registerInventory()
    {
        BloodMagicAPI.INSTANCE.registerInventoryProvider("curiosInventory", player -> getCuriosInventory(player));
    }

    public NonNullList<ItemStack> getCuriosInventory(Player player)
    {
        Optional<IItemHandler> optionalItemHandler = CuriosApi.getCuriosHelper().getEquippedCurios(player).resolve();
        NonNullList<ItemStack> inventory = NonNullList.create();

        if (optionalItemHandler.isPresent())
        {
            IItemHandler itemHandler = optionalItemHandler.get();
[...]

PR #2045, opened 2024-06-13
fix log spam on death

walksanatora:
closes #2038 by making function return a empty list if the curios list is not present

PR #2139, opened 2025-12-11
Fixed curio related logspam

Saereth:
#2085 and #2040
 Added null-safety checks:
  1. Check if curioInv.isEmpty() before calling .get()
  2. Check if livingArmourSockets == null before using it

### DIFF
--- a/src/main/java/wayoftime/bloodmagic/compat/CuriosCompat.java
+++ b/src/main/java/wayoftime/bloodmagic/compat/CuriosCompat.java
@@ -40,7 +40,14 @@ public class CuriosCompat
 
 	public int recalculateCuriosSlots(Player player)
 	{
-		ICurioStacksHandler livingArmourSockets = CuriosApi.getCuriosInventory(player).resolve().get().getCurios().get("living_armour_socket");
+		Optional<ICuriosItemHandler> curioInv = CuriosApi.getCuriosInventory(player).resolve();
+		if (curioInv.isEmpty()) {
+			return 0;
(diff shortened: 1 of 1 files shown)


wrincewind:
I've had a look and am unable to replicate this issue in live - i'm not seeing logspam on death, or when wearing living armour, or when playing in a small pack wth Corail Tombstone. Were you able to replicate this issue before deploying your fix, and if so, how? 

As far as i know this was fixed by #2045, for reference.

Saereth:
The Problem is the original code - assumes everything exists
  `ICurioStacksHandler livingArmourSockets =    CuriosApi.getCuriosInventory(player).resolve().get().getCurios().get("living_armour_socket");`

This chain fails if:
- resolve() returns an empty Optional -> .get() throws NoSuchElementException
- The "living_armour_socket" slot doesn't exist -> returns null, then next operation NPEs
  `Optional<ICuriosItemHandler> curioInv = CuriosApi.getCuriosInventory(player).resolve();`

`if (curioInv.isEmpty()) { return 0; }`  // Guard #1
  ICurioStacksHandler livingArmourSockets = curioInv.get().getCurios().get("living_armour_socket");`


`if (livingArmourSockets == null) { return 0; }`  // Guard #2

Why it helps:
- Returns early with 0 instead of crashing/spamming logs
- Handles edge cases like Curios not being fully initialized, or the slot type not being registered yet, or having been removed. 

Hope that clears it up, happy to elaborate more if needed

Issue #2030, opened 2024-03-15
[Error] Can't place player // Serverside exception // Related to spectral tiles


MaxWasUnavailable:
#### Issue Description:
Players randomly can't join the server until (sometimes) a restart of the server resolves it. 

Relevant debug log snippet:

```
[15Mar2024 17:10:53.394] [Server thread/ERROR] [net.minecraft.server.network.ServerLoginPacketListenerImpl/]: Couldn't place player in world
java.lang.NullPointerException: Cannot invoke "net.minecraft.world.level.block.state.BlockState.m_60734_()" because "p_129203_" is null
        at net.minecraft.nbt.NbtUtils.m_129202_(NbtUtils.java:254) ~[server-1.20.1-20230612.114412-srg.jar%23785!/:?]
        at wayoftime.bloodmagic.common.tile.TileSpectral.serialize(TileSpectral.java:82) ~[bloodmagic-1.20.1-3.3.2-44.jar%23492!/:1.20.1-3.3.2-44]
        at wayoftime.bloodmagic.common.tile.base.TileBase.m_5995_(TileBase.java:132) ~[bloodmagic-1.20.1-3.3.2-44.jar%23492!/:1.20.1-3.3.2-44]
        at net.minecraft.network.protocol.game.ClientboundLevelChunkPacketData$BlockEntityInfo.m_195691_(ClientboundLevelChunkPacketData.java:152) ~[server-1.20.1-20230612.114412-srg.jar%23785!/:?]
        at net.minecraft.network.protocol.game.ClientboundLevelChunkPacketData.<init>(ClientboundLevelChunkPacketData.java:44) ~[server-1.20.1-20230612.114412-srg.jar%23785!/:?]
        at net.minecraft.network.protocol.game.ClientboundLevelChunkWithLightPacket.<init>(ClientboundLevelChunkWithLightPacket.java:24) ~[server-1.20.1-20230612.114412-srg.jar%23785!/:?]
        at net.minecraft.server.level.ChunkMap.m_183760_(ChunkMap.java:1261) ~[server-1.20.1-20230612.114412-srg.jar%23785!/:?]
        at net.minecraft.server.level.ChunkMap.m_183754_(ChunkMap.java:892) ~[server-1.20.1-20230612.114412-srg.jar%23785!/:?]
[...]

Issue #2026, opened 2024-02-17
TileSpectral


Reaviik:
#### Issue Description:
Note: If this bug occurs in a modpack, please report this to the modpack author. Otherwise, delete this line and add your description here. If this is a feature request, this template does not apply to you. Just delete everything.

#### Steps to reproduce:

1. It happens occasionally, as I own a server and several players reported this error to me, I bring it here. Not even WorldEdit can replace the block. however, this block appears in places where players had not yet started playing blood magic

2. This server is running Arclight version arclight-1.20.1-1.0.4-fd335c1 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT)
...

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: 1.20.1-3.3.2-44
- Minecraft: 1.20.1
- Forge: 47.0.20

Reaviik:
Sorry about my English. Is there any way to remove this block through the mod? or get his coordinates

Reaviik:
I got a command from ATM's redit to remove the blocks, from the sigil of suppression
/fill ~-10 ~-10 ~-10 ~25 ~14 ~25 minecraft:air replace bloodmagic:spectral
apparently happens when there are more than 1000 blocks

gamerpotion:
We are having similar issue with this block on our server. It corrupted earlier and was restricting players from logging on. Please could this be fixed?   Also that block has missing texture fyi.

<img src="https://media.discordapp.net/attachments/964187610392199209/1211372217770836008/image.png?ex=65edf54a&is=65db804a&hm=1ca65bcfaa842c726ffaa21476b0339cfe3b9a715d098e999551be570862f64c&=&format=webp&quality=lossless">

<img src="https://media.discordapp.net/attachments/964187610392199209/1211373058296782919/Screenshot_2024-02-25-20-02-44-052_com.miui.gallery-edit.jpg?ex=65edf613&is=65db8113&hm=4204f7d33af8776baf1a523d0e57f293d742916f97acb164cb3c6eb61bc8a3df&=&format=webp&width=1920&height=421">

<img src="https://media.discordapp.net/attachments/964187610392199209/1211373946314952735/2024-02-25_18.07.23.png?ex=65edf6e7&is=65db81e7&hm=7c53a4a7d6622f420a68f11b4b1c116601a249f8045001410a26bb5078b7e736&=&format=webp&quality=lossless&width=1702&height=671">

[...]

PR #2140, opened 2025-12-11
The TileSpectral fix for issue #2030

Saereth:
#2030 and #2026
Prevents NPE which causes players not to be able to connect


wrincewind:
I've been trying to replicate this bug in the latest version, but i haven't been able to. As far as i was aware it was fixed by #2096 - if it still exists in the most recent release, can you tell me how you were able to replicate it?

Saereth:
It was partially (mostly fixed) there is still edge case scenarios where the stored block can become null or invalid (mod removals etc) this is just a guard against the edge case to prevent an NPE.

PR #2142, opened 2025-12-14
fix up altar checking

stellanera98:
doesnt depend on patchouli anymore, make datapacked structures more restrictive but easier to create and understand while still following the general "rules" of the altar structure

also moving rune upgrades to capabilities (fueled by data map for the vanilla ones) so animus could still do the variable rune amount thing (which I only learned about being a thing recently) and allow for other (and more) dynamic rune kinds


stellanera98:
@dphaldes

src/main/java/wayoftime/bloodmagic/common/caps/BMCaps.java
```diff
@@ -0,0 +1,8 @@
+package wayoftime.bloodmagic.common.caps;
+
+import net.neoforged.neoforge.capabilities.BlockCapability;
+import wayoftime.bloodmagic.BloodMagic;
+
+public class BMCaps {
+    public static final BlockCapability<IBloodRune, Void> BLOOD_RUNE = BlockCapability.createVoid(BloodMagic.rl("blood_rune"), IBloodRune.class);
```
dphaldes: `BLOOD_RUNE` doesnt convey what it does. this could be something like `RUNE_POWER` or `RUNE STRENGTH`. anything that conveys that this is capability is linked with runes' property

src/main/java/wayoftime/bloodmagic/common/caps/IRunePowers.java
dphaldes, line 1: Could use a better name for the Interface and function name. getUpgrades() works but maybe there is something better

dphaldes (approved):
Other than some name changes, this looks like a solid improvement!

src/main/java/wayoftime/bloodmagic/common/caps/BMCaps.java
stellanera98: its not only the power/strength but also the kind/type of rune. Best I can come up with right now would be UPGRADE_PROVIDER and IUpgradeProvider#provideUpgrades but Im not really sure this is any better.

In 1.20 there were separate calls for the type and amount, but that restricts to a single type per block which I wanted to avoid ~~so I can add a capacity/acceleration hybrid to not "waste" spots on a reinforced altar~~

dphaldes (approved):

PR #2143, opened 2025-12-14
Imperfect Rituals

stellanera98:
- datamap controls which blocks cause which effects
- effects are configured in a datapack registry
- effect types can be added in code


stellanera98:
@dphaldes

src/main/resources/data/bloodmagic/loot_table/test.json
dphaldes, line 1: is this test part of the PR / something we want to retain ?

stellanera98: I used it to make sure the loot table override works, its probably not needed anymore but I figured Ill keep it for now rather than recreating it later for something else

dphaldes (approved):
Already looked at the API the other day. Looks good!

PR #2144, opened 2025-12-15
Sigils

stellanera98:
- costs and effect are defined in datapack
- additional effect kinds can be registered in code
- there is only one sigil item, putting the sigil_type data component on it (ResourceKey<SigilType>) will determine its behaviour
- an item model for the sigil has to be placed at `assets/<type modid>/models/item/sigil_<type path>`
- translation keys `item.bloodmagic.sigil.<type path>` (item name) and `tooltip.bloodmagic.sigil.<type path>` are pretty much required
- if the sigil is a toggleable one (like fast miner or green growth) it needs the sigil_active data component set. if its not set the sigil will not toggle it on shift + use
- the sigil_active data component is also used as an item model override returning 1 for true and 0 for false, to be used on toggleable sigils

@dphaldes


dphaldes:
nitpick: 
> data/bloodmagic/bloodmagic/sigil_type

I wonder if this path be renamed to just "sigils" type doesnt really add much

src/main/java/wayoftime/bloodmagic/api/sigil/SigilEffect.java
```diff
@@ -0,0 +1,68 @@
...
+            .getRegistry()
+            .get()
+            .byNameCodec()
+            .dispatch(SigilEffect::codec, Function.identity())
+    );
+
+    Supplier<MapCodec<ApplyPotionEffect>> APPLY_POTION_EFFECT = SIGIL_EFFECT_TYPE.register("apply_potion_effect", () -> ApplyPotionEffect.CODEC);
+    Supplier<MapCodec<DivinationEffect>> DIVINATION_EFFECT = SIGIL_EFFECT_TYPE.register("divination_effect", () -> DivinationEffect.CODEC);
+    Supplier<MapCodec<FluidPlaceEffect>> FLUID_PLACE_EFFECT = SIGIL_EFFECT_TYPE.register("fluid_place_effect", () -> FluidPlaceEffect.CODEC);
+    Supplier<MapCodec<FluidRemoveEffect>> FLUID_REMOVE_EFFECT = SIGIL_EFFECT_TYPE.register("fluid_remove_effect", () -> FluidRemoveEffect.CODEC);
```
dphaldes: Extract all the implementations and the map into a separate class in common instead of API.
keep the interface bare with only Sigil related content

src/main/java/wayoftime/bloodmagic/api/sigil/SigilType.java
```diff
@@ -0,0 +1,14 @@
+package wayoftime.bloodmagic.api.sigil;
+
+import com.mojang.serialization.Codec;
+import com.mojang.serialization.codecs.RecordCodecBuilder;
+
+public record SigilType(int airCost, int blockCost, int entityCost, int refreshCost, SigilEffect effect) { // TODO arcane ash sigil beacon thingy?
+    public static final Codec<SigilType> CODEC = RecordCodecBuilder.create(builder -> builder.group(
+            Codec.INT.fieldOf("click_air_cost").forGetter(SigilType::airCost),
+            Codec.INT.fieldOf("click_block_cost").forGetter(SigilType::blockCost),
+            Codec.INT.fieldOf("click_entity_cost").forGetter(SigilType::entityCost),
```
dphaldes: I would call them "use_empty_cost", "use_block_cost" and "use_entity_cost" respectively

dphaldes:
> if the sigil is a toggleable one (like fast miner or green growth) it needs the sigil_active data component set. if its not set the sigil will not toggle it on shift + use

This should be done with the datapack directly. Add a field to the sigil types something along the line of "activatable" and use that to add and remove ACTIVE datacomponent on use. That way new activateable sigils are purely data pack driven and wont need the additional component set on them by default

src/main/java/wayoftime/bloodmagic/common/creativetab/BMTabs.java
```diff
@@ -53,6 +59,25 @@ public class BMTabs {
...
+            ItemStack tmp = sigilStack.copy();
+            tmp.set(BMDataComponents.SIGIL_TYPE, key);
+            tab.accept(tmp);
+        });
+        List<ResourceKey<SigilType>> toggle = List.of(Sigils.MINER);
+        toggle.forEach(key -> {
+            ItemStack tmp = sigilStack.copy();
+            tmp.set(BMDataComponents.SIGIL_TYPE, key);
+            tmp.set(BMDataComponents.SIGIL_ACTIVE, false);
+            tab.accept(tmp);
```
dphaldes: not needed if you use my suggestion

src/main/java/wayoftime/bloodmagic/common/datacomponent/BMDataComponents.java
```diff
@@ -29,6 +32,9 @@ public class BMDataComponents {
     public static final DeferredHolder<DataComponentType<?>, DataComponentType<Double>> DEMON_WILL_AMOUNT = DATA_COMPONENTS.registerComponentType("will_amount", builder -> builder.persistent(Codec.DOUBLE).networkSynchronized(ByteBufCodecs.DOUBLE));
     public static final DeferredHolder<DataComponentType<?>, DataComponentType<EnumWillType>> DEMON_WILL_TYPE = DATA_COMPONENTS.registerComponentType("will_type", builder -> builder.persistent(EnumWillType.CODEC).networkSynchronized(EnumWillType.STREAM_CODEC));
 
+    public static final DeferredHolder<DataComponentType<?>, DataComponentType<Boolean>> SIGIL_ACTIVE = DATA_COMPONENTS.registerComponentType("sigil_active", builder -> builder.persistent(Codec.BOOL).networkSynchronized(ByteBufCodecs.BOOL));
```
dphaldes: Can be made in a Unit type if you use my suggestion

dphaldes (approved):
Implementation looks solid. some small changes with names and suggestion

src/main/java/wayoftime/bloodmagic/api/sigil/SigilEffect.java
stellanera98: so the effect and registry registering goes in common/sigil/SigilEffects and the interface itself stays in API. Codec definition stays in interface or would you put that with the other stuff?

src/main/java/wayoftime/bloodmagic/api/sigil/SigilType.java
stellanera98: imo empty is weird there but use seems better than click

stellanera98:
putting the activatable in datapack seems better, Ill get to that

src/main/java/wayoftime/bloodmagic/api/sigil/SigilEffect.java
dphaldes: Codec can stay there. Extract the implementations and registry out

stellanera98:
using Unit for the active type makes checks on if its active shorter, but toggling it turns from setting !state to an if/else which seems a bit excessive. I do like the boolean approach more but its not a big difference

src/main/java/wayoftime/bloodmagic/common/item/SigilItem.java
dphaldes, line 1: Are you sure we want this file in API?

dphaldes (approved):
Looks great!

src/main/java/wayoftime/bloodmagic/common/item/SigilItem.java
stellanera98: It shouldn't. When I moved it I was still mentally with the implementation version that required it to make new sigils, but thats not how it works anymore

stellanera98:
- moved SigilItem back to common/items
- removed SigilType
- SigilEffect methods now return an int, this is then used as the cost
- SigilEffect now has `String getActiveTooltip(boolean isActive)` which by default returns translation keys for "Activated" and "Deactivated" but could be used to say something else where appropriate (see discord thread about it regarding the Exchange Sigil)
  - there has to be a better name for this


@dphaldes

src/main/java/wayoftime/bloodmagic/common/sigil/ApplyPotionEffect.java
```diff
@@ -0,0 +1,61 @@
...
+
+public record ApplyPotionEffect(Holder<MobEffect> effect, int selfAmplifier, int selfTicks, int upkeep, Optional<Integer> otherAmplifier, Optional<Integer> otherTicks, Optional<Integer> otherCost) implements SigilEffect {
+    public static final MapCodec<ApplyPotionEffect> CODEC = RecordCodecBuilder.mapCodec(builder -> builder.group(
+            MobEffect.CODEC.fieldOf("effect_type").forGetter(ApplyPotionEffect::effect),
+            Codec.INT.fieldOf("amplifier").forGetter(ApplyPotionEffect::selfAmplifier),
+            Codec.INT.fieldOf("duration").forGetter(ApplyPotionEffect::selfTicks),
+            Codec.INT.fieldOf("upkeep").forGetter(ApplyPotionEffect::upkeep),
+            Codec.INT.optionalFieldOf("amplifier_others").forGetter(ApplyPotionEffect::otherAmplifier),
+            Codec.INT.optionalFieldOf("duration_others").forGetter(ApplyPotionEffect::otherTicks),
+            Codec.INT.optionalFieldOf("cost_others").forGetter(ApplyPotionEffect::otherCost)
```
dphaldes: What do these fields do? the keys in the datapack are no immediately clear

src/main/java/wayoftime/bloodmagic/common/sigil/ApplyPotionEffect.java
```diff
@@ -0,0 +1,61 @@
...
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.Level;
+import wayoftime.bloodmagic.api.sigil.SigilEffect;
+
+import java.util.Optional;
+
+public record ApplyPotionEffect(Holder<MobEffect> effect, int selfAmplifier, int selfTicks, int upkeep, Optional<Integer> otherAmplifier, Optional<Integer> otherTicks, Optional<Integer> otherCost) implements SigilEffect {
+    public static final MapCodec<ApplyPotionEffect> CODEC = RecordCodecBuilder.mapCodec(builder -> builder.group(
+            MobEffect.CODEC.fieldOf("effect_type").forGetter(ApplyPotionEffect::effect),
```
dphaldes: I would call this "potion_effect_type" or "mob_effect_type" explicitly so that we dont get confused by Sigil Effects or any other effect

src/main/java/wayoftime/bloodmagic/common/sigil/ApplyPotionEffect.java
```diff
@@ -0,0 +1,61 @@
...
+            Codec.INT.fieldOf("amplifier").forGetter(ApplyPotionEffect::selfAmplifier),
+            Codec.INT.fieldOf("duration").forGetter(ApplyPotionEffect::selfTicks),
+            Codec.INT.fieldOf("upkeep").forGetter(ApplyPotionEffect::upkeep),
```
dphaldes: if "upkeep" is a cost, lets add "_cost" suffix to it just in case

src/main/java/wayoftime/bloodmagic/common/sigil/FluidPlaceEffect.java
```diff
@@ -0,0 +1,50 @@
...
+import net.neoforged.neoforge.fluids.FluidType;
+import net.neoforged.neoforge.fluids.FluidUtil;
+import net.neoforged.neoforge.fluids.capability.IFluidHandler;
+import wayoftime.bloodmagic.api.sigil.SigilEffect;
+
+public record FluidPlaceEffect(Holder<Fluid> fluid, int tankFillAmount, int cost) implements SigilEffect {
+    public static final MapCodec<FluidPlaceEffect> CODEC = RecordCodecBuilder.mapCodec(builder -> builder.group(
+            BuiltInRegistries.FLUID.holderByNameCodec().fieldOf("fluid_type").forGetter(FluidPlaceEffect::fluid),
+            Codec.INT.fieldOf("tank_fill_amount").forGetter(FluidPlaceEffect::tankFillAmount),
+            Codec.INT.fieldOf("success_cost").forGetter(FluidPlaceEffect::cost)
```
dphaldes: can be just "cost", success is redundant

src/main/java/wayoftime/bloodmagic/common/sigil/FluidPlaceEffect.java
```diff
@@ -0,0 +1,50 @@
...
+import net.neoforged.neoforge.fluids.FluidStack;
+import net.neoforged.neoforge.fluids.FluidType;
+import net.neoforged.neoforge.fluids.FluidUtil;
+import net.neoforged.neoforge.fluids.capability.IFluidHandler;
+import wayoftime.bloodmagic.api.sigil.SigilEffect;
+
+public record FluidPlaceEffect(Holder<Fluid> fluid, int tankFillAmount, int cost) implements SigilEffect {
+    public static final MapCodec<FluidPlaceEffect> CODEC = RecordCodecBuilder.mapCodec(builder -> builder.group(
+            BuiltInRegistries.FLUID.holderByNameCodec().fieldOf("fluid_type").forGetter(FluidPlaceEffect::fluid),
+            Codec.INT.fieldOf("tank_fill_amount").forGetter(FluidPlaceEffect::tankFillAmount),
```
dphaldes: should probably call it just "fluid_amount"  and use the same term in FluidRemoveEffect too

src/main/java/wayoftime/bloodmagic/common/sigil/ApplyPotionEffect.java
stellanera98: Nothing atm, they are intended for the beacon like effect with arrays

stellanera98: To specify, in 1.20 code the not fully implemented beacon like array effect with this sigil would have given the minecraft:haste effect for 600 ticks and an amplifier of 2 (so Haste III ingame) for 600 ticks for 1 HP damage (100 essence by default conversion) dealt to all the players in its AoE that receive it. I included those so it'd be easier later. See also SigilEffect#hasArrayEffect() and SigilEffect#arrayTick(...)

dphaldes (approved):

dphaldes (approved):

Issue #2146, opened 2025-12-16
Clearing flight and suspended effects through /clear or milk does not properly get rid of the effect of the effects


305pence:
Steps to reproduce:
1. give yourself a milk bucket
2. go to survival
3. give yourself the flight effect (through commands or potion)
4. fly
5. drink the milk
6. you will not fall to the ground, you will also be able to turn off and on the flight as if the effect never ended. 

this persists after relog and can be fixed if you drink another flight potion and let the effect run out

for blood magic version 3.3.5 minecraft version 1.20.1

stellanera98:
Flight is enabled and disabled on MobEffectEvent.Added and .Expired, Id assume the expired one doesn't get fired when cleared. If theres events for clearing/removing as well thats an easy fix, otherwise the method of giving flight needs to be changed

PR #2147, opened 2025-12-16
Remove fly ability on potion effect clear as well

stellanera98:
#2146 
MobEffectEvent.Remove exists and is fired on /effect clear and using a milk bucket


dphaldes:
Instead of this, we should use a custom MobEffect implementation and override `removeAttributeModifiers` method from where you can reset the ability

stellanera98:
If there is an attribute for flight already, sure

dphaldes:
There isnt an attribute, you still need an ability but that method gives you a callback to do it cleanly instead of relying on events and checking for potion effect.

stellanera98:
Ah, Ill have a look at that in a bit then, working on sigils atm

stellanera98:
@dphaldes seems to work

dphaldes (approved):

PR #2148, opened 2025-12-16
Soft Touch & Fortuna Extract fix try 2

stellanera98:
Tested soft touch on shulker box with items in it, multiple times worked every time, same items still in it
Tested soft touch on amethyst cluster, got some drops of 4 but also a bunch with more


Saereth:
While this will mostly work you're still working with a stale be, so there is some room for edge cases here fwiw but this approach is basically the adhoc implementation so it should work.. mostly...

wrincewind (approved):

PR #2152, opened 2025-12-18
Fix NullPointerException in getTier method and add error handling for altar tier and rune upgrades

OmiLabsDev:


dphaldes:
@stellanera98

stellanera98:
Im afraid this is irrelevant after #2142

stellanera98:
Also tier value -1 means something went so wrong it *should* crash, which it does because in the currently merged code it doesn't load the structures from the datapack registry as the code for that is commented out and had been waiting to be replaced by the pr mentioned above. 

Should probably check for it and throw an error that says what the actual issue is instead, but ignoring it is definitely not the way to go

Issue #2110, opened 2025-10-26
Missing Mixins


bookerthegeek:
## Heyo

Saw this in my log, thought you might like to know. I was not in a development environment.

``` json5
[main/WARN] [mixin/]: Reference map 'bloodmagic.refmap.json' for bloodmagic.mixins.json could not be read. If this is a development environment you can ignore this message

```

### Versions
 - Minecraft: 1.20.1
 - Forge: 47.4.10
 - BloodMagic: 3.3.3-45

PR #2129, opened 2025-11-28
Reconfigured the jar and jarjar task for dep inclusion

Saereth:
Reconfigured the 'jar' and 'jarJar' tasks to clearly define
their roles. The default jar will now include bundled dependencies,
while the slim jar will exclude them, indicated by the 'slim' classifier.

PR #2155, opened 2025-12-20
1.20 mixin

stellanera98:
not sure it was necessary in the end since at least part of the issue was not having it reobfuscate the jarJar file, but:
- made MixinLivingEntity extend Entity
- removed MixinEntity
this is because A) the shadowed method is already public and B) https://wiki.fabricmc.net/tutorial:mixin_tips suggests extending the parent class of the target in such cases
- suffix for default (non jarJar) jar is now -slim
- jarJar task now also gets the manifest thingy that jar had. To be clear, I do not know where this data is going or if its needed
- jarJar file is now included in the "publish" gradle task
- trying to remove the "slim" jar made gradle unhappy because of implicit dependencies, dont know why and dont want to deep dive into forgegradle or the publish plugin to find out, it may stay ~~and sit in a dusty corner where it'll be ignored~~

Based on #2129 with permission of the author


stellanera98:
@dphaldes

src/main/java/wayoftime/bloodmagic/mixin/client/MixinLivingEntity.java
```diff
@@ -2,15 +2,23 @@
...
 
 @Mixin(LivingEntity.class)
-public abstract class MixinLivingEntity extends MixinEntity
+public abstract class MixinLivingEntity extends Entity
 {
+    public MixinLivingEntity(EntityType<?> p_19870_, Level p_19871_) {
+        super(p_19870_, p_19871_);
+        throw new IllegalStateException("Instantiated MixinLivingEntity");
+    }
+
```
dphaldes, line 21: Why is this needed ?  It's an abstract class and won't be instantiated

wrincewind:
will this fix #2110 btw?

stellanera98:
Im pretty sure that warning and the spectral sight not working are caused by the same issue that is resolved by this pr, yes

wrincewind:
All right, great, and what about @dphaldes comment?

wrincewind (approved):

PR #2159, opened 2025-12-24
annoying item interactions

stellanera98:
- Teleposition sigil didnt let you teleport when clicking any block, instead of just resetting its target on teleposers
- ritual diviner and tinkerer didnt let you cycle through their things when looking at a block
- Teleposition sigil didnt work properly when inside of a sigil of holding
- Holding sigil didnt care about the result from the contained sigils use method (but did for useOn)

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -29,6 +29,8 @@ AgriCraft is *not* installed.
 - prevent a crash when opening the Edit HUD position screen while looking at an entity
 - soft coating no longer voids the contents of shulker boxes (or other tile-entities)
 - bloodmagic:geode_harvestable now adds forge:clusters as a tag instead of a block, meaning it'll actually work with them
+- teleposition sigil now teleports you when clicking on a non-teleposer block as well. it also now checks itself for a destination when inside a sigil of holding
+- ritual diviner and tinkerer can now be cycled even when looking at a (non master ritual stone) block
 
 ------------------------------------------------------
 Version 3.3.5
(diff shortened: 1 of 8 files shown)


wrincewind:
would you mind removing 'while looking at empty air' from https://github.com/WayofTime/BloodMagic/blob/1.20.1/src/main/resources/assets/bloodmagic/patchouli_books/guide/en_us/entries/rituals/ritual_diviner.json#L8 as part of this pr? Oh, and add a note to the changelog too, if you'd be so kind. I've checked the other manual pages bu tthis is the only place this issue was mentioned.

stellanera98:
I've put it in the changelog txt because I thought you might want to write the ingame one yourself, but I can add that. I'll also have the other thing removed from the book

wrincewind:
I'm not going to be able to do any dev work until *maybe* the 27th/28th but more likely not until the 3rd-5th, so i'm happy for you to update the changelog in the manual - i'd do it myself but can't see where the run-offs are :p 

also, i can't see changelog.txt in your PR?

stellanera98:
Huh? I know I wrote a bit about it and Im pretty sure that was *before* I ran git add... Probably gonna do a clean install of dumbJ when I get home, been having weird issues with it everywhere.
Just sad java doesn't have a functional standalone language server

stellanera98:
actually might have been the geode bit I added earlier. that *was* in the .txt

wrincewind (approved):

PR #2160, opened 2025-12-24
Move stuff to API

stellanera98:
moved stuff I think should be in /api there

not sure how to do something similar with:
- attributes (self sacrifice boost)
- data component types (binding, arc tool stats, living armour stuff, etc)
- capabilities (rune stats)
- data attachments? (not sure if needed but incense)

I also think at least parts of LivingHelper should be in api, but there stuff thats just there to make the effects work probably arent among them. Potentially split into "make it work" and "stuff you might wanna call from 'outside'".

Another thing is items/blocks etc. might be nice to have access to those in api so add-ons dont need to go look up the name. Potentially put the resource locations for those somewhere in api (and then use those to give the name to the deferred register? sounds halfway stupid to me but its an idea)

@dphaldes (less about code since its just renaming and moving stuff, more about the things mentioned above)


src/main/java/wayoftime/bloodmagic/api/datacomponent/Binding.java
dphaldes, line 1: I wonder if this needs more suitable class name

src/main/java/wayoftime/bloodmagic/api/helper/RitualUtil.java
dphaldes, line 1: Dont think this package belongs in API.

src/main/java/wayoftime/bloodmagic/common/damagesource/BMDamageSources.java
dphaldes, line 1: What happened to this class?

stellanera98: they arent actually the sources but the damage types, renamed because of it and moved to api

src/main/java/wayoftime/bloodmagic/api/helper/RitualUtil.java
stellanera98: it might once the "real" rituals are done. might be a premature move though

dphaldes: Exposing Util classes to API might not be a great idea. We probably don't want people to poke around it

src/main/java/wayoftime/bloodmagic/api/datacomponent/Binding.java
stellanera98: SoulNetworkBinding, SoulNetworkLink, BoundPlayer

src/main/java/wayoftime/bloodmagic/api/helper/RitualUtil.java
stellanera98: I was thinking it might be useful for things that are common in "ritual effects" to go there so addons can make use of it. the lightning strike, creating a predictable fake player for rituals doing stuff, interacting with ranges (though that should be on the range classes directly I think)
Any tailored thing would go with the ritual effect it belongs to directly

stellanera98:
should be ready to merge now

Issue #1862, opened 2021-10-25
Water sigil destroys blocks


jeremiahwinsley:
#### Issue Description:
When using a water sigil on a partial block, like stair/slab/fence, instead of waterlogging the block it places the water next to it, destroying anything in the way.

#### What happens:

https://user-images.githubusercontent.com/7946299/138631804-1c94fb29-7efb-4b10-88cc-daca17daefe0.mp4

#### Steps to reproduce:

1. Place fence surrounded by bedrock.
2. Right click water sigil on the side of the fence.

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: 3.1.5-25
- Minecraft: 1.16.5
- Forge: 36.2.8

PR #2167, opened 2026-02-22
add check for empty block in fluid sigil placement, and waterlogging

Phil-M-H:
1. Fixes Issue [1862](https://github.com/WayofTime/BloodMagic/issues/1862)
2. Added waterlogging to the water sigil

I edited the changelog as well since I saw some other PRs do that -- just let me know if you'd prefer I don't touch that.


Phil-M-H:
To be clear, issue #1862 still exists in the latest versions (I accidentally broke an end portal frame on a server I was hosting because I right-clicked on the side of an eye of ender), I just dug up the issue so it could be closed out.

stellanera98:
I've been adding to the changelog because its usually just me and thus everything is in order and github shouldnt be whining about there being mismatches. also theres the changelog.txt and the changelog book entry which should both be updated at the same time.

It would be easier for us if you removed the changelog entry and we'll add it ourselves once we merge it

wrincewind (approved):

Issue #1880, opened 2022-01-25
Tick accelerating a Crafting Alchemy array results in duplicate outputs


MuteTiefling:
#### Issue Description:
Tick accelerating an activated Alchemy Array causes the output items to be duplicated

#### What happens:
The faster the tick acceleration, the more items are generated

#### What you expected to happen:
Only one output per craft.

#### Steps to reproduce:

1. Put down an alchemy array and put the ingredients in it, 
2. As the animation starts, accelerate it using Time in a Bottle
3. The faster you make it go the more items will be duplicated

____
#### Affected Versions (Do *not* use "latest"):

- BloodMagic: BloodMagic-1.16.4-3.1.7-27
- Minecraft: 1.16.5
- Forge: 36.2.22

https://github.com/EnigmaticaModpacks/Enigmatica6/issues/4130

PR #2168, opened 2026-04-02
Prevent tick acceleration from duping items in the alchemy array

Saereth:
return false from attemptCraft() when the craft completes, so tick() takes the else branch and resets state immediately


stellanera98:
looking at the code this should work as intended

you should probably remove the fixes thingy since we want to keep the issue open but tagged as "resolved in upcoming version" until the version containing the fix is released. I think we had this before but it would have been a while and Im not 100% sure about it

wrincewind:
What was the issue # that this fixes, again? Just for my notes.

stellanera98:
#1880

wrincewind (approved):

PR #2171, opened 2026-04-14
Alchemy Table

stellanera98:
- alchemy table
- basic recipe type/serializer/builder
- gui/menu


wrincewind (approved):
note to self, test recipes when some are added (particularly, check the gui flags for "missing orb", "orb not bound" "orb not high enough tier" and "not enough LP")

PR #2181, opened 2026-05-02
Formatters

stellanera98:


wrincewind:
this is deeply cursed, but to be fair, you're fixing my cock-up, so cheers :p

Issue #2177, opened 2026-04-22
Spec Proposal: Rich Will-Type Tooltips for Sentient Weapons & Tools


Saereth:
The Animus bloodmagic addon adds an enhanced tooltip to the Sentient Bow (ItemSentientBow#appendHoverText) that goes beyond the minimal currentType line already present on upstream sentient items. 

## Scope

Apply the expanded tooltip pattern to:
  - ItemSentientSword
  - ItemSentientAxe
  - ItemSentientPickaxe
  - ItemSentientShovel
  - ItemSentientScythe

  Each currently shows only a flavour line + tooltip.bloodmagic.currentType.<type>.

  Proposed Tooltip Contents

  Per item, in order:

  1. Flavour line: existing tooltip.bloodmagic.<item>.desc (unchanged, italic gray).
  2. Attuned will type: highlighted (color depending on type), ie. "Attuned: Destructive". Promotes the existing currentType line to a more scannable color.
  3. Will level + pool: computed client-side from the player's will of the current type: "Level 3 (240 will)".
  4. Per-level effect readout: the item's headline scaling stat at the current level:
    - Sword/Axe/Pickaxe/Shovel: bonus damage
    - Scythe: bonus damage + AoE radius (scythe already tracks both)
    - Will type specific riders (for example, Vengeful lifesteal %, Corrosive poison duration) behind a "hold Shift" expand.

wrincewind:
I'd suggest making it so that the attuned will type's text is coloured based on the will that it's attuned to - e.g. Corrosive is green, Destructive is yellow, etc. Otherwise yeah, this all sounds great.

Saereth:
updated to reflect text color based on will type

PR #2182, opened 2026-05-08
Add detailed tooltips for Sentient items

Saereth:
These tooltips now display current Sentient level, Demon Will pool, bonus damage, and specific rider effects (such as, Wither duration, Absorption time, Movement Speed). Detailed stats are shown when holding Shift.

completes #2177 

image

---

image


src/main/java/wayoftime/bloodmagic/common/item/soul/SentientTooltipHelper.java
```diff
@@ -0,0 +1,140 @@
...
+			break;
+		case STEADFAST:
+			tooltip.add(Component.translatable("tooltip.bloodmagic.sentient.rider.steadfast",
+					formatNumber(steadfastAbsorptionTimeTicks / 20.0)).withStyle(color));
+			break;
+		case VENGEFUL:
+			if (vengefulMovementSpeed > 0)
+			{
+				tooltip.add(Component.translatable("tooltip.bloodmagic.sentient.rider.vengeful",
+						formatPercent(vengefulMovementSpeed)).withStyle(color));
```
stellanera98: the speed granted is flat, not %

src/main/java/wayoftime/bloodmagic/common/item/soul/SentientTooltipHelper.java
```diff
@@ -0,0 +1,140 @@
...
+			s = s.replaceAll("0+$", "");
+			if (s.endsWith("."))
+				s = s.substring(0, s.length() - 1);
+		}
+		return s;
+	}
+
+	private static String formatSigned(double value)
+	{
+		String sign = value >= 0 ? "+" : "";
```
stellanera98, line 132: shouldnt this be `0 ? "+" : "-"`?
to be clear Im not entirly sure there is a negative number anywhere relevant, but if there was it should be noted as such

stellanera98 (changes requested):
the rest of it looks good to me

src/main/java/wayoftime/bloodmagic/common/item/soul/SentientTooltipHelper.java
Saereth: Yeah all the values fed in (bonusDamage, digSpeedBonus) are non-negative but the existing logic handles negatives correctly regardless here. formatNumber delegates to String.format() which already prefixes - for negatives so the function only needs to prepend "+" for non-negatives.

src/main/java/wayoftime/bloodmagic/common/item/soul/SentientTooltipHelper.java
Saereth: Ah yeah I always forgot that was weirdly a flat increase, sorted

stellanera98 (approved):

